PR #15, opened 2018-10-02
Print strat

Reidmcc:
Add current strategy to the log for later analysis. Change is to address issue https://github.com/lightyeario/kelp/issues/8


cmd/trade.go
```diff
@@ -70,6 +70,25 @@ func init() {
 		}
 		log.Printf("Trading %s:%s for %s:%s\n", botConfig.ASSET_CODE_A, botConfig.ISSUER_A, botConfig.ASSET_CODE_B, botConfig.ISSUER_B)
 
+		//Add current strategy to the log
+		var strat_display string
+		switch *strategy {
```
nikhilsaraf: We can do without the switch-case to keep it simple. Just use *strategy instead directly in the log statement.

Reidmcc:
I've made the requested change, makes sense.

PR #16, opened 2018-10-02
doc clarify full path needed for configs

Reidmcc:
Edit trade examples to clarify that the full path to the file is required.

The examples at present don't make it clear that the path to the _.cfg_ file is required for the config flags. This made it difficult for me as a new user to get started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,13 +107,13 @@ These are the following commands available from the `kelp` binary:
 
 The `trade` command has three parameters which are:
 
-- **botConf**: _.cfg_ file with the account details, [sample file here](examples/configs/trader/sample_trader.cfg).
+- **botConf**: full path to the _.cfg_ file with the account details, [sample file here](examples/configs/trader/sample_trader.cfg).
 - **strategy**: the strategy you want to run (_sell_, _buysell_, _balanced_, _mirror_, _delete_).
-- **stratConf**: _.cfg_ file specific to your chosen strategy, [sample files here](examples/configs/trader/).
+- **stratConf**: full path to the _.cfg_ file specific to your chosen strategy, [sample files here](examples/configs/trader/).
 
 Here's an example of how to start the trading bot with the _buysell_ strategy:
 
-`kelp trade --botConf trader.cfg --strategy buysell --stratConf buysell.cfg`
+`kelp trade --botConf ./path/trader.cfg --strategy buysell --stratConf ./path/buysell.cfg`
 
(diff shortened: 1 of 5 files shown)


examples/walkthroughs/trader/buysell.md
```diff
@@ -41,9 +41,9 @@ A level defines a [layer](https://en.wikipedia.org/wiki/Layering_(finance)) that
 
 ## Run Kelp
 
-Assuming your botConfig is called `trader.cfg` and your strategy config is called `buysell.cfg`, you can run `kelp` with the following command:
```
nikhilsaraf: let's keep the walkthrough files (`buysell.md`, `sell.md`, `balanced.md`) consistent w.r.t. keeping/removing the following text:
```
you can run `kelp` with the following command
```

your PR only removed it from this file

README.md
```diff
@@ -107,13 +107,13 @@ These are the following commands available from the `kelp` binary:
...
-- **botConf**: _.cfg_ file with the account details, [sample file here](examples/configs/trader/sample_trader.cfg).
+- **botConf**: full path to the _.cfg_ file with the account details, [sample file here](examples/configs/trader/sample_trader.cfg).
 - **strategy**: the strategy you want to run (_sell_, _buysell_, _balanced_, _mirror_, _delete_).
-- **stratConf**: _.cfg_ file specific to your chosen strategy, [sample files here](examples/configs/trader/).
+- **stratConf**: full path to the _.cfg_ file specific to your chosen strategy, [sample files here](examples/configs/trader/).
 
 Here's an example of how to start the trading bot with the _buysell_ strategy:
 
-`kelp trade --botConf trader.cfg --strategy buysell --stratConf buysell.cfg`
+`kelp trade --botConf ./path/trader.cfg --strategy buysell --stratConf ./path/buysell.cfg`
```
nikhilsaraf: The addition of `./path/` seems intuitive. Thanks for adding.
Let's keep the rest of the sample commands similar, i.e. replace `.path/` with `./path/` in the files below

examples/walkthroughs/trader/buysell.md
Reidmcc: That removal was supposed to be a revert to the original text when I simplified my change. I'll put it back here instead of removing it from the others.

Reidmcc:
I've implemented your consistency suggestions (thanks for the catch) and put the phrase about 'you can run....' back in the one file where I removed it.

PR #21, opened 2018-10-04
Place windows build correctly

Reidmcc:
Rework build.sh to address problems with Windows builds. This should finish resolving https://github.com/lightyeario/kelp/issues/14

- Made the `need to invoke from the root 'kelp' directory` check conditional on not being in Windows because the `if [[ ($# -gt 1 || `pwd | rev | cut -d'/' -f1 | rev` != "kelp") ]]` condition always fails in Windows
- Changed the `OUTFILE` to explicitly write to `$GOPATH/bin/`. This fixes the misplacement issue.
- Removed `mkdir -p bin`. All it seems to do is create `kelp/bin`, at least in Windows. It shouldn't be necessary in any case with `$GOPATH/bin/` explicitly set.
- Added a text log output. In Windows the script opens a new window command window that immediately closes when complete, preventing the user from seeing the output if it completes quickly, as it did with the `need to invoke from the root 'kelp' directory` error. This helps with diagnostics. 
- I'm not sure why this pull request is showing my commits from yesterday

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: go
+
+go_import_path: github.com/lightyeario/kelp
+
+go:
+- "1.10.x"
+- "1.11"
+
+before_install:
+- curl https://glide.sh/get | sh
+
+install:
+- glide install
+
+script:
+- ./scripts/build.sh
+- ./bin/kelp version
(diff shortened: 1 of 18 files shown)


nikhilsaraf:
thanks for this PR, will look at it by the end of this week, I'm just getting a windows VM set up to test it.

Reidmcc:
Cool, thanks. I tested it on the same machine as https://github.com/lightyeario/kelp/issues/14, in Windows 10

nikhilsaraf:
@Reidmcc I tested this on a Windows VM and it worked correctly when running under a `bash` shell. Once you install `bash` it should just work! That is the expected way to run this on windows and I've added a note in the README to indicate this.

Closing this PR along with the original issue. If you can't get the build script to work for you still then let me know. thanks for taking the time to look into this.

PR #22, opened 2018-10-04
Fix divide by zero error

Reidmcc:
Fix the divide by zero error from https://github.com/lightyeario/kelp/issues/5

Added a conditional to check that both relevant numbers are >0. This should exclude cases where no valid number exists as well. I think this would need to be done for all _UpdateWithOps_ funcs across strategy plugins.

### DIFF
--- a/plugins/sellSideStrategy.go
+++ b/plugins/sellSideStrategy.go
@@ -1,8 +1,8 @@
 package plugins
 
 import (
+	"fmt"
 	"log"
-	"math"
 
 	"github.com/lightyeario/kelp/api"
 	"github.com/lightyeario/kelp/model"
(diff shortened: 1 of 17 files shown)


nikhilsaraf:
wanted to acknowledge this PR, will look at it soon.

Reidmcc:
Cool, thanks.

plugins/sellSideStrategy.go
```diff
@@ -111,28 +111,33 @@ func (s *sellSideStrategy) UpdateWithOps(offers []horizon.Offer) (ops []build.Tr
 		// hitCapacityLimit can be updated below
 		targetPrice := s.currentLevels[i].Price
 		targetAmount := s.currentLevels[i].Amount
-		if s.divideAmountByPrice {
-			targetAmount = *model.NumberFromFloat(targetAmount.AsFloat()/targetPrice.AsFloat(), targetAmount.Precision())
-		}
+		if targetAmount.AsFloat() > 0 && targetPrice.AsFloat() > 0 {
```
nikhilsaraf: the code convention is to return errors when they are encountered rather than to wrap the real logic in deeply nested if blocks (you can imagine how deep this can get if you have a series of error checks).

instead of using the pattern:

```
    if (condition) {
        // do something
    } else {
        // log error
    }
```

can you use this pattern:

```
    if targetPrice.AsFloat() == 0 {
        return nil, nil, fmt.Errorf("targetPrice is 0")
    }
    if targetAmount.AsFloat() == 0 {
        return nil, nil, fmt.Errorf("targetAmount is 0")
    }
    // rest of the code stays here
```

note that in the suggestion above we're returning an error rather than handling it by logging. the caller should handle any errors that arise from this method.

Reidmcc:
Changes made. Once we have this finalized and committed I'll check the rest of the _UpdateWithOps_ funcs.

PR #32, opened 2018-10-23
Change variables for config structs from ALL_CAPS to CamelCase

Reidmcc:
Change variables for config structs from ALL_CAPS to CamelCase. 

This should resolve https://github.com/lightyeario/kelp/issues/6

### DIFF
--- a/plugins/buysellStrategy.go
+++ b/plugins/buysellStrategy.go
@@ -17,7 +17,7 @@ type buySellConfig struct {
 	RateOffsetPercentFirst bool          `valid:"-" toml:"RATE_OFFSET_PERCENT_FIRST"`
 	AmountOfABase          float64       `valid:"-" toml:"AMOUNT_OF_A_BASE"` // the size of order to keep on either side
 	DataTypeA              string        `valid:"-" toml:"DATA_TYPE_A"`
-	DataTypeAURL           string        `valid:"-" toml:"DATA_FEED_A_URL"`
+	DataFeedAURL           string        `valid:"-" toml:"DATA_FEED_A_URL"`
 	DataTypeB              string        `valid:"-" toml:"DATA_TYPE_B"`
 	DataFeedBURL           string        `valid:"-" toml:"DATA_FEED_B_URL"`
 	Levels                 []staticLevel `valid:"-" toml:"LEVELS"`
@@ -42,7 +42,7 @@ func makeBuySellStrategy(
 	}
 	sellSideFeedPair, e := MakeFeedPair(
 		config.DataTypeA,
-		config.DataTypeAURL,
+		config.DataFeedAURL,
 		config.DataTypeB,
 		config.DataFeedBURL,
 	)
@@ -74,7 +74,7 @@ func makeBuySellStrategy(
 		config.DataTypeB,
 		config.DataFeedBURL,
 		config.DataTypeA,
-		config.DataTypeAURL,
+		config.DataFeedAURL,
 	)
 	if e != nil {
 		return nil, fmt.Errorf("cannot make the buysell strategy because we could not make the buy side feed pair: %s", e)
--- a/plugins/buysellStrategy.go
+++ b/plugins/buysellStrategy.go
@@ -1,6 +1,8 @@
 package plugins
 
 import (
+	"fmt"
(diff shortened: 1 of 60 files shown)


nikhilsaraf:
looks good at the first glance, will dig deeper tomorrow.

You'll need to fix the `StructString()` function in `support/utils/configs.go` to use the correct display name for the config so the configs are displayed with the ALL_CAPS version of the toml string.

You can use the following code snippet in line 30 of `support/utils/configs.go`:

		field := reflect.TypeOf(s).Field(i)
		fieldName := field.Name
		fieldDisplayName := field.Tag.Get("toml")
		if fieldDisplayName == "" {
			fieldDisplayName = fieldName
		}

Reidmcc:
Yep, I had missed those files. (don't mind the weird commit list, just an issue with my local branch)

plugins/buysellStrategy.go
```diff
@@ -10,17 +10,17 @@ import (
...
-	DATA_FEED_B_URL           string        `valid:"-"`
-	LEVELS                    []staticLevel `valid:"-"`
+	PriceTolerance         float64       `valid:"-" toml:"PRICE_TOLERANCE"`
+	AmountTolerance        float64       `valid:"-" toml:"AMOUNT_TOLERANCE"`
+	RateOffsetPercent      float64       `valid:"-" toml:"RATE_OFFSET_PERCENT"`
+	RateOffset             float64       `valid:"-" toml:"RATE_OFFSET"`
+	RateOffsetPercentFirst bool          `valid:"-" toml:"RATE_OFFSET_PERCENT_FIRST"`
+	AmountOfABase          float64       `valid:"-" toml:"AMOUNT_OF_A_BASE"` // the size of order to keep on either side
+	DataTypeA              string        `valid:"-" toml:"DATA_TYPE_A"`
+	DataTypeAURL           string        `valid:"-" toml:"DATA_FEED_A_URL"`
```
nikhilsaraf: typo: `DataTypeAURL` should be `DataFeedAURL`

terminator/config.go
```diff
@@ -21,15 +21,15 @@ type Config struct {
 // String impl.
 func (c Config) String() string {
 	return utils.StructString(c, map[string]func(interface{}) interface{}{
-		"SOURCE_SECRET_SEED":  utils.SecretKey2PublicKey,
-		"TRADING_SECRET_SEED": utils.SecretKey2PublicKey,
+		"SourceSecretSeed":  utils.SecretKey2PublicKey,
```
nikhilsaraf: If we were to run terminator currently then it would log out the secret keys because this transformation would not run.

Since `StructString` changed to take in a map keyed by the `feedDisplayName` this should be changed back to `SOURCE_SECRET_SEED` and `TRADING_SECRET_SEED` instead of using variable names.
This also applies to the `String()` method in `trader/config.go` which currently runs because the map it uses is keyed by the correct values (and is inconsistent with this method).

I'm fine with all these functions using either the _variableName_ or the _CONFIG_NAME_ as long as they are all consistent and run the transformations correctly.

support/utils/configs.go
```diff
@@ -27,18 +27,23 @@ func StructString(s interface{}, transforms map[string]func(interface{}) interfa
...
+		fieldName := field.Name
+		fieldDisplayName := field.Tag.Get("toml")
+		if fieldDisplayName == "" {
+			fieldDisplayName = fieldName
+		}
 
 		// set the transformation function
 		transformFn := passthrough
-		if fn, ok := transforms[fieldName]; ok {
+		if fn, ok := transforms[fieldDisplayName]; ok {
```
nikhilsaraf: using `fieldName` here requires the `transforms` map to use the _variableName_, whereas using `fieldDisplayName` requires the map to use the _CONFIG_NAME_.
see related comment in `terminator/config.go`.

nikhilsaraf:
@Reidmcc the updates looks good.

There’s a conflict in `trader/config.go` — there was a new config param that was added. Could you resolve that conflict by converting it to the new camelCased format — this PR can be merged once that’s resolved. Thanks!

Reidmcc:
I've updated the alert variables, should be good to go.

Issue #30, opened 2018-10-19
Write log lines to a file with log rotation


nikhilsaraf:
## Desired Behavior
I want to be able to save log entries in between runs of kelp automatically

## Impact
The desired behavior will allow me to retrospectively look back at the bot's actions without having to manually rotate log files

## Feature Suggestion
We can achieve the desired behavior by outputting log lines directly to a file rather than to stdout that has a date suffix in the name

## References
This is a common feature in almost all applications

Issue #9, opened 2018-09-10
Prettier logs


nikhilsaraf:
## Desired Behavior
I want the logs to be more intuitive and easy to read.

## Impact
The desired behavior will allow me to more easily gauge what the bot is doing and will also help with retrospective inspection of specific strategies.

## Feature Suggestion
We can achieve the desired behavior by:

1. having structured log lines at the `sdex.go` level that are well-documented.
2. using a logging framework that supports colors: https://github.com/op/go-logging

Reidmcc:
I could have a go at this one. Before I get started, what level of generalization are you thinking of? 

I'm envisioning something like functions that format and return messages in a particular category in a consistent logging framework format. For example all log messages that are errors would be the same color and have the same metadata elements. Is that the right track?

Reidmcc:
Ok, sounds good. So right now there are just these steps:

1. Set up the logger interface
2. Switch all the log messages to use that interface
3. Make sure it can still output to file

And specifically *not* to create any complex loggers like the web logger example.

nikhilsaraf:
yes, you got all that correct.

on the output to file, that will still use the current logic that you had added previously (i.e. we will not have a separate fileLogger and regular logger for the first implementation, just a `LogLogger` implementation).

jmg421:
I know nothing about how to do this sort of thing but am happy to help.  I have kelp v1.4.0 installed on my Mac.

nikhilsaraf:
@jmg421 You should compile the latest version of Kelp from source, see the README for how to do this along with setting up the necessary tools. I'd recommend creating a new branch in your own fork for this project.

You can then start refactoring the log lines one file at a time from `log.Printf` to `l.Infof` and `l.Errorf`. See a sample commit that does this [here](https://github.com/stellar/kelp/commit/7366a03cc7fa14ed3fa49d9a02e28e17a5a7fcc8#diff-e22974e446f0478d01339ac5f579dc3d) for the `cmd/trade.go file. You can then submit PRs against those changes to be included in the master branch. Please keep your Pull Requests to be for 1 PR per file. This will allow the code reviews to go quicker.

Let me know if you have any questions along the way!

PR #33, opened 2018-10-23
Add file logging

Reidmcc:
Add logging to time-stamped auto-generated file

Resolves https://github.com/lightyeario/kelp/issues/30

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -14,8 +14,8 @@ import (
 	"github.com/stellar/go/support/config"
 )
 
-const tradeExamples = `  kelp trade --botConf trader.cfg --strategy buysell --stratConf buysell.cfg
-  kelp trade --botConf trader.cfg --strategy buysell --stratConf buysell.cfg --sim`
+const tradeExamples = `  kelp trade --botConf ./path/trader.cfg --strategy buysell --stratConf ./path/buysell.cfg
+  kelp trade --botConf ./path/trader.cfg --strategy buysell --stratConf ./path/buysell.cfg --sim`
 
 var tradeCmd = &cobra.Command{
 	Use:     "trade",
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -68,7 +68,7 @@ func init() {
 
 		if *logPrefix != "" {
 			t := time.Now().Format("20060102T150405MST")
-			fileName := fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", *logPrefix, botConfig.ASSET_CODE_A, botConfig.ISSUER_A, botConfig.ASSET_CODE_B, botConfig.ISSUER_B, t)
+			fileName := fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", *logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, t)
 			e := setLogFile(fileName)
 			if e != nil {
 				log.Println()
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -66,8 +66,6 @@ func init() {
 		}
 
 		if *logPrefix != "" {
-			//t := time.Now().String()
-			//test_20181024T154304PDT
 			t := time.Now().Format("20060102T150405MST")
 			fileName := fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", *logPrefix, botConfig.ASSET_CODE_A, botConfig.ISSUER_A, botConfig.ASSET_CODE_B, botConfig.ISSUER_B, t)
 			e := setLogFile(fileName)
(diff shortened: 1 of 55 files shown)


cmd/trade.go
```diff
@@ -188,3 +193,18 @@ func deleteAllOffersAndExit(botConfig trader.BotConfig, client *horizon.Client,
 		log.Fatal("...nothing to delete, exiting")
 	}
 }
+
+func setLogFile() {
```
nikhilsaraf: we should pass in the filename var to this function

the filename `prefix` should be taken in as an optional input from the command line. If nothing is specified then it should default to std.out, i.e. it should not call the `setLogFile()` function in that case.

before calling the `setLogFile` function you can append the `timeString` (`t`) to create the `filename` passed into the function.

cmd/trade.go
```diff
@@ -188,3 +193,18 @@ func deleteAllOffersAndExit(botConfig trader.BotConfig, client *horizon.Client,
...
+	t := time.Now().String()
+	var b strings.Builder
+	b.WriteString("KelpLog_")
```
nikhilsaraf: don't worry about using string builder etc. -- can just use `fileName := fmt.Sprintf("%s_%s", prefix, t)`, which should be in the calling method, based on the comment above

Reidmcc:
Ok, makes sense to do it as part of https://github.com/lightyeario/kelp/issues/9. 

Shall I close out this pull request or leave it in case we want it later?

nikhilsaraf:
Doing it as part of issue #30 is correct since that deals with both writing to a log file and adding log rotation etc. #9 is meant only to have prettier logging.

if you can make the changes from the comments then I'll try to merge this in as a partial solution to #30. We can think about log rotation and truncation later -- I may have something that can be used to solve that easily. Making it easy for Kelp to write to a file is more important in the short term than supporting that with rotation and truncation.

Reidmcc:
Ok, I've reworked it to use a `--log` flag on the end of the trade command

cmd/trade.go
```diff
@@ -64,6 +62,13 @@ func init() {
 		}
 		log.Println(startupMessage)
 
+		if *logPrefix != "" {
+			t := time.Now().String()
```
nikhilsaraf: the time string produces a weird long filename with spaces. example:
`test_2018-10-24 15:43:04.414657477 -0700 PDT m=+0.003531672`

this should produce a timestamp in your local timezone formatted like so (year, month, day, T, hour, min, second, tz):
`test_20181024T154304PDT`

see the [Time.Format](https://golang.org/pkg/time/#Time.Format) function in the _time_ package for reference

cmd/trade.go
```diff
@@ -64,6 +62,13 @@ func init() {
 		}
 		log.Println(startupMessage)
 
+		if *logPrefix != "" {
```
nikhilsaraf: this should be the first thing we do, even before writing the starting message (above). This way the startup message, which includes the version and git hash, will also be a part of the log file.

cmd/trade.go
```diff
@@ -194,12 +199,7 @@ func deleteAllOffersAndExit(botConfig trader.BotConfig, client *horizon.Client,
...
-func setLogFile() {
-	t := time.Now().String()
-	var b strings.Builder
-	b.WriteString("KelpLog_")
-	b.WriteString(t)
-	fileName := b.String()
+func setLogFile(fileName string) {
 	f, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
 	if err != nil {
 		log.Fatalf("Failed to set log file: %v", err)
```
nikhilsaraf: this should not log a fatal error here but should wrap the error and return it, example:

    return fmt.Errorf("failed to set log file: %s", e)

the calling code can then check for that error and log a Fatal error like so:

 	if e != nil {
 	 	log.Println()
 	 	log.Fatal(e)
 	 	return
 	 }

also, in Kelp the convention is to use `e` to denote errors instead of using `err`. it's shorter and much easier to read.

cmd/trade.go
nikhilsaraf: can you also give the filename a `.log` extension

Reidmcc: I implemented this without making the issuers explicitly optional to avoid bulky conditional code, same approach as the later `"Trading %s:%s for %s:%s\n"` message. This results in a double underscore when XLM is an asset , i.e `XLM__BTC` but I don't think that's a problem.

Reidmcc:
I implemented the new file format without making the issuers explicitly optional to avoid bulky conditional code, same approach as the later `"Trading %s:%s for %s:%s\n"` message. This results in a double underscore when XLM is an asset , i.e `XLM__BTC` but I don't think that's a problem.

cmd/trade.go
```diff
@@ -67,6 +64,27 @@ func init() {
 			log.Println()
 			log.Fatal(e)
 		}
+
+		if *logPrefix != "" {
+			//t := time.Now().String()
```
nikhilsaraf: Can you remove the commented out code and sample string?

Reidmcc: Yep, meant to do that.

Reidmcc:
Commented code and sample removed.

PR #37, opened 2018-10-27
Add new strategy trackSDEX

Reidmcc:
### Strategy description
This strategy allows Kelp to track the SDEX orderbook price as the midpoint for setting levels, and allows the user to manually set the amount per level. It dynamically adjusts level amount to maintain asset balance close to value parity.

The strategy assumes that the SDEX price is correct and is thus is most appropriate for native SDEX tokens.

I'll write up a walkthrough and add to the README once the code is finalized.

### New files

The added files follow the convention of the existing strategies:

- Base strategy plugin:  **trackSDEXStrategy.go**
- Level provider plugin: **trackSDEXLevelProvider.go**
- Price feed plugin:       **sdexFeed.go**
- Configuration file example: **sample_trackSDEX.cfg**

### Modifications to existing files

Functions necessary to enable the strategy were added to two files:

- Added the strategy's `StrategyContainer` to **factory.go**
- Added a function to pull SDEX orderbooks to **functions.go**


Reidmcc:
The minimum total order check making the strategy maker-only partially fulfills https://github.com/lightyeario/kelp/issues/34, but only for this strategy. To fully resolve https://github.com/lightyeario/kelp/issues/34 we'll have to integrate a generalized maker-only function and add the command parameter.

Reidmcc:
The minimum total order code did not perform as intended; in fact it made the bot cross itself. Reverted to original raw midpoint checking.

Reidmcc:
**Problem**: Since the Price Tolerance/Amount Tolerance check is done against existing levels, this strategy's instruction to not place orders, and thus generate no levels, on one side if that side's asset is too low makes the bot not track changes in that direction after the first time it places no orders on that side. So the orders on the other side will sit passively, regardless of market activity in the opposite direction (behavior observed in live testing). 

Possible solutions:

- Incorporate this strategy's `MAINTAIN_BALANCE_PERCENT` parameter into the amount checks in `PreUpdate`. I don't think we'd want to change the strategy-external code for specific strategies though, and I'm not seeing an obvious place to modify capacity (e.g. to change `maxAssetBase` or `maxAssetQuote`) in strategy-internal functions before `PreUpdate`.
- Change this strategy's logic such that it generates a single level with a very small amount instead of 0 levels in the case of low asset balance so the bot has something to evaluate. This is an inelegant work-around, but I think it would work.

Reidmcc:
Closing this pull request due to it getting really messy when I pulled in updates from root. Will re-open.

PR #39, opened 2018-10-30
Add trackSDEX strategy

Reidmcc:
### Strategy description

This strategy allows Kelp to track the SDEX orderbook price as the midpoint for setting levels, and allows the user to manually set the amount per level. It dynamically adjusts level amount to maintain asset balance close to value parity.

The strategy assumes that the SDEX price is correct and is thus is most appropriate for native SDEX tokens.

I'll write up a walkthrough and add to the README once the code is finalized.

### New files

The added files follow the convention of the existing strategies:

- Base strategy plugin: trackSDEXStrategy.go
- Level provider plugin: trackSDEXLevelProvider.go
- Price feed plugin: sdexFeed.go
- Configuration file example: sample_trackSDEX.cfg

### Modifications to existing files

Functions necessary to enable the strategy were added to two files:

Added the strategy's StrategyContainer to factory.go
Added a function to pull SDEX orderbooks to functions.go

### DIFF
--- a/examples/configs/trader/sample_trackSDEX.cfg
+++ b/examples/configs/trader/sample_trackSDEX.cfg
@@ -1,0 +1,25 @@
+# Sample config file for the "trackSDEX" strategy
+
+# what % deviation from the ideal price is allowed before we reset the price, specified as a decimal (0 < PRICE_TOLERANCE < 1.00)
+PRICE_TOLERANCE=0.01
+
+# what % deviation from the ideal amount is allowed before we reset the price, specified as a decimal (0 < AMOUNT_TOLERANCE < 1.00)
+AMOUNT_TOLERANCE=0.01
+
+# define the bid/ask spread that you are willing to provide. spread is a percentage specified as a decimal number (0 < spread < 1.00) - here it is 0.1%
+SPREAD=0.002
+
+# What base percent of your (one sided) balance to use on each level specified as a decimal number
+BASE_PERCENT_PER_LEVEL = 0.01
+
+# max number of levels to have on either side. Defines how deep of an orderbook you want to make.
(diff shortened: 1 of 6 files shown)


nikhilsaraf:
@Reidmcc the strategy in this PR seems more targeted. It may be better to keep it in your own fork for now.

- - - 

I will be adding better support for custom, pluggable, and shareable strategies hopefully by end of year. These will be run in a REST wrapper that Kelp can interact with, similar to how we interact with CCXT via the ccxt-rest process.

I’m thinking of the best way to share these strategies that are created by the community with the community (reddit, keybase, slack, website, registry file in the kelp repo, etc.). If you have any suggestions please let me know. Maybe your strategy can be the first one?

- - - 

The sdex price feed seems generalizable — could you include that in the list of price feeds by extending the current set of tags (crypto, fixed, exchange, etc.) to include one called “sdex”? The URL format can be like **”code1:issuer1/code2:issuer2”** where we can leave an empty issuer for XLM markets: **”XLM:/BTC:GA....END”**.

Reidmcc:
@nikhilsaraf I'm having trouble getting the SDEX feed integrated with the existing `FeedPair` type. In order to get the SDEX orderbook I need to pass in the `*SDEX` client construct from `makeBuySellStrategy`, but `MakeFeedPair` only accommodates text. 

I'd just have `makeBuySellStrategy` grab the SDEX price from the bot's trading assets (like I did in my strategy), but `staticSpreadLevelProvider` expects `*api.FeedPair`. I'm not finding a way around it.

Alternately we could implement this as a separate SDEX tracking strategy, with the buysell logic instead of the custom logic I have in this pull request.

Reidmcc:
Closing this; the code is too old to merge reasonably and the custom strategy stuff isn't going in. A clean sdex price feed PR is at https://github.com/interstellar/kelp/pull/90

Issue #52, opened 2018-11-13
Prevent balanced strategy from modifiying orders when nothing has changed


Reidmcc:
## Desired Behavior

Currently, if the balanced strategy has _LEVEL_DENSITY_ < 1.0 it will modify its orders to re-randomize the levels even if no offers have been taken.

I want to have the balanced strategy leave all levels in place if no trades have happened since the last refresh; I don't think re-randomizing is useful when nothing has changed.

## Impact

The current behavior submits unnecessary operations to the network. This change will increase efficiency and reduce user fee spending.

## Feature Suggestion

We can implement this by having the bot stop modifying offers if the first offer is still in place and within tolerances. I'm submitting a pull request shortly.

nikhilsaraf:
@Reidmcc this is a good idea.
The change should probably be made in balancedLevelProvider.go

balancedLevelProvider can save (in memory) it's asset balances and the levels it returns from the GetLevels call. Anytime we call GetLevels it can first check if the asset balances have changed from what it has saved in memory:
- if it has changed then it can recompute new levels
- otherwise it can return the levels it has saved in memory.

This will be a pretty quick and simple code change.

The suggested approach will not consider amount tolerances since any balance change will trigger levels to be recomputed. Adjusting this to handle amount tolerances can be worked on as a separate issue which is a relatively small improvement over the suggested approach.

Reidmcc:
That make sense. Do you have the code change planned out or shall I take a crack at it?

nikhilsaraf:
@Reidmcc The approach described above is what I had been thinking about for a while but never got to making the change. Would be great if you could take a crack at it!

Reidmcc:
Ok, will do.

nikhilsaraf:
@Reidmcc I've uploaded the fill tracking and handling logic, you can take a look at how that works for the mirror strategy as an example: https://github.com/interstellar/kelp/blob/master/plugins/mirrorStrategy.go#L312

PR #53, opened 2018-11-13
Stop sellSideStrategy from modifiying unchanged states

Reidmcc:
This pull request addresses https://github.com/lightyeario/kelp/issues/52. By having the bot check explicitly whether a level was actually modified we can implement logic based on that status, in this case having the bot stop modifying if the first level is unchanged.


nikhilsaraf:
@Reidmcc we can always have the first level not be modified but a subsequent level be modified.
think of the example where your level provider randomly returns one of these two results:

    1 - 0.25, 0.27
    2 - 0.25, 0.29

In the above example, the first level would not be modified but the subsequent levels would be modified.

Reidmcc:
Closing this pull request; the are better ways to do this than hard-coding it into sellSideStrategy.

nikhilsaraf:
moving the conversation over to the issue you opened #52

PR #61, opened 2018-11-19
Fix crossover bug

Reidmcc:
This should address https://github.com/lightyeario/kelp/issues/4 using the proposed fix of resetting the price if the price trigger didn't fire. I didn't live-replicate the bug because I don't have enough tokens to reliably get a partially filled level :-)


Reidmcc:
makes sense; this way is more of an emergency stop

Reidmcc:
closing this pr; we will not use this approach

PR #71, opened 2018-12-02
Add loggers package, update trade.go

Reidmcc:
Here's the foundation of the new loggers package and the sample update to trade.go.

This is the start of implementation for https://github.com/interstellar/kelp/issues/9.


support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
+package loggers // I didn't call the packacke "logging" because the go-loggeing framework's package is "logging" so we'll need to be different when we switch
```
nikhilsaraf: in that case we can call it logger (drop the `s`)

support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"log"
```
nikhilsaraf: we should not import `log` here.
this file should only contain the Logger interface.

BasicLogger should be in it's own file `support/logger/basicLogger.go` or `support/logger/nativeLogger.go`

support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// added Fatal and Fatalf because trade.go (and elsewhere) use log.Fatal
+	// fatal error messages, with newline
+	Fatal(e error)
+
+	// formatted fatal error messages, without automatic newline
+	Fatalf(e error, args ...interface{})
+}
+
+type BasicLogger struct {
```
nikhilsaraf: exported types should be commented -- are you using [go-metalinter](https://github.com/alecthomas/gometalinter) in your IDE?

I think for now BasicLogger can be unexported (`basicLogger` instead of `BasicLogger`) -- it should have a factory method `MakeBasicLogger` which is itself exported and returns a `Logger` (see example of usage in the issue)

support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (l BasicLogger) Info(msg string) {
```
nikhilsaraf: you should also add the line that enforces that BasicLogger implements a Logger:
```golang
var _ Logger = &BasicLogger{}
```

and also use pointers for the receivers of the methods, example: `func (l *BasicLogger) Info(msg string)`

support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+	// basic messages, can be custom formatted, similar to fmt.Printf. User needs to add a \n if they want a newline after the log entry
+	Infof(msg string, args ...interface{})
+
+	// error messages, indicates to the logger that these messages can be handled differently (different color, special format, email alerts, etc.). The type of logger will determine what to do with these messages. The logger should NOT panic on these messages. Appends a newline (\n) after each entry.
+	Error(msg string)
+
+	// error messages, indicates to the logger that these messages can be handled differently (different color, special format, email alerts, etc.). The type of logger will determine what to do with these messages. The logger should NOT panic on these messages. User needs to add a \n if they want a newline after the log entry.
+	Errorf(msg string, args ...interface{})
+
+	// added Fatal and Fatalf because trade.go (and elsewhere) use log.Fatal
```
nikhilsaraf: I'd rather that we remove `Fatal` and `Fatalf` as methods on the interface for now -- we can use `os.exit()` in the places where we call `log.Fatal` (very few, only at the top level). This will reduce the burden on the log implementations to have to implement these additional two methods which are essentially just an `Error` followed by an `os.exit()`

we should add a method here to help with that:
```golang
// Fatal is a convenience method that can be used with any Logger to log a fatal error
func Fatal(l Logger, e error) {
    l.Info("")
    l.Error(e)
    os.exit(1)
}
```

this will be used like so:
```golang
// some code that gives us an e that we want to log as a fatal error
logger.Fatal(l, e)
```

support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type BasicLogger struct {
+}
+
+func (l BasicLogger) Info(msg string) {
+	log.Println(msg)
+}
+
+func (l BasicLogger) Infof(msg string, args ...interface{}) {
```
nikhilsaraf: all these methods should have comments based on the go format since they are exported (see other files in the project). you can just keep it simple for the basic logger like `// Infof impl` since it's not doing anything complex.

support/loggers/loggers.go
```diff
@@ -0,0 +1,53 @@
...
+	log.Printf(msg, args...)
+}
+
+func (l BasicLogger) Error(msg string) {
+	log.Println(msg) // isn't actually differnt from Info until we do structured logs
```
nikhilsaraf: we can remove these comments -- structured logs will be a whole new logger implementation

cmd/trade.go
```diff
@@ -31,6 +32,8 @@ var tradeCmd = &cobra.Command{
 	Example: tradeExamples,
 }
 
+var logger loggers.BasicLogger
```
nikhilsaraf: this should not be a top level variable like this.

the logger should be created inside the function `tradeCmd.Run = func(ccmd *cobra.Command, args []string)`. Eventually the config files will dictate where we want to log which will guide the logger object that is created.

we can start the convention in kelp that the logger variable will always be called `l` (lowercase-L). so we can always log using `l.Infof(...)` as an example. Any objects that themselves hold a logger can use it in their own functions like so: `obj.l.Infof()` instead of `obj.logger.Infof()` or `obj.log.Infof()`. This convention is akin to using `e` for errors in kelp instead of `err` to keep things more succinct.

cmd/trade.go
```diff
@@ -93,20 +96,20 @@ func init() {
 		utils.CheckConfigError(botConfig, e, *botConfigPath)
 		e = botConfig.Init()
 		if e != nil {
-			log.Println()
-			log.Fatal(e)
+			logger.Info("")
+			logger.Fatal(e)
```
nikhilsaraf: this will become:
```golang
logger.Fatal(l, e)
```

see comments below for more info

support/loggers/loggers.go
Reidmcc: No, I hadn't been using metalinter. I assume you recommend it?

Reidmcc:
I rebased this branch since so much had changed...yeah bad plan. I'm putting in a new pull request.

PR #78, opened 2018-12-05
Add balanced fill tracking to avoid unecessary re-randomization

Reidmcc:
This implements a fill tracker for the balanced strategy that simply toggles a boolean for whether to regen levels. If no fills happen the existing levels will be re-submitted.

Addresses https://github.com/interstellar/kelp/issues/52

### DIFF
--- a/plugins/balancedLevelProvider.go
+++ b/plugins/balancedLevelProvider.go
@@ -1,6 +1,7 @@
 package plugins
 
 import (
+	"fmt"
 	"log"
 	"math/rand"
 	"time"
@@ -27,7 +28,7 @@ type balancedLevelProvider struct {
 	virtualBalanceBase            float64 // virtual balance to use so we can smoothen out the curve
 	virtualBalanceQuote           float64 // virtual balance to use so we can smoothen out the curve
 	orderConstraints              *model.OrderConstraints
-	needNewLevels                 bool // boolean for whether to generate levels, starts true
+	shouldRefresh                 bool // boolean for whether to generate levels, starts true
 
 	// precomputed before construction
 	randGen *rand.Rand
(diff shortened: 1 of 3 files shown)


plugins/balancedLevelProvider.go
```diff
@@ -27,14 +27,21 @@ type balancedLevelProvider struct {
 	virtualBalanceBase            float64 // virtual balance to use so we can smoothen out the curve
 	virtualBalanceQuote           float64 // virtual balance to use so we can smoothen out the curve
 	orderConstraints              *model.OrderConstraints
+	needNewLevels                 bool // boolean for whether to generate levels, starts true
```
nikhilsaraf: can use a more terse name: `shouldRefresh`

plugins/balancedLevelProvider.go
```diff
@@ -95,35 +105,19 @@ func validateSpread(spread float64) {
...
+		log.Println("no offers were taken, leave levels as they are")
+		return p.lastLevels, nil
+	}
 
-		// always take a spread off the amountCarryover
-		amountCarryoverSpread := p.getRandomSpread(p.minAmountCarryoverSpread, p.maxAmountCarryoverSpread)
-		amountCarryover *= (1 - amountCarryoverSpread)
+	levels, e := p.recomputeLevels(maxAssetBase, maxAssetQuote)
+	if e != nil {
+		log.Fatalf("unable to generate new levels: %s\n", e)
```
nikhilsaraf: we should **never** use log.Fatal
instead we should wrap the error like this: `fmt.Errorf("unable to generate new levels: %s", e)`.
The framework will handle retries etc.

Reidmcc:
@nikhilsaraf updates made

nikhilsaraf:
@Reidmcc made some updates and merging.

PR #88, opened 2019-01-10
Logger framework and trade.go example

Reidmcc:
This is a replacement for https://github.com/interstellar/kelp/pull/71, incorporating requested changes.


Reidmcc:
I left the logger variable as a top-level declaration in trade.go because if we move it into `tradeCmd.Run = func(ccmd *cobra.Command, args []string)` the logger is unavailable to the other top level functions, e.g. `logPanic()`. Alternately we could change the other functions to have them return their messages instead of logging the messages inside the functions but that's a bigger rewrite. Which solution do you prefer?

cmd/trade.go
```diff
@@ -31,6 +32,8 @@ var tradeCmd = &cobra.Command{
 	Example: tradeExamples,
 }
 
+var l *logger.BasicLogger
```
nikhilsaraf: we should not have *any* global vars -- if you need this in the logPanic() method then logPanic can take in the logger interface

cmd/trade.go
```diff
@@ -48,7 +51,8 @@ func hiddenFlag(flag string) {
 func logPanic() {
 	if r := recover(); r != nil {
 		st := debug.Stack()
-		log.Printf("PANIC!! recovered to log it in the file\npanic: %v\n\n%s\n", r, string(st))
+		l.Errorf("PANIC!! recovered to log it in the file\npanic: %v\n\n%s\n", r, string(st))
+		os.Exit(1)
```
nikhilsaraf: why exit here?

cmd/trade.go
```diff
@@ -210,8 +215,8 @@ func init() {
 
 		strategyFillHandlers, e := strat.GetFillHandlers()
 		if e != nil {
-			log.Println()
-			log.Printf("problem encountered while instantiating the fill tracker: %s\n", e)
+			l.Info("")
+			l.Infof("problem encountered while instantiating the fill tracker: %s\n", e)
```
nikhilsaraf: this should follow the same pattern as the logging of the monitoring server to keep things consistent. either change both to a new pattern or none.

cmd/trade.go
```diff
@@ -308,49 +312,48 @@ func validateTrustlines(client *horizon.Client, botConfig *trader.BotConfig) {
...
 		e := sdex.SubmitOps(dOps, func(hash string, e error) {
 			if e != nil {
-				log.Println()
-				log.Fatal(e)
+				logger.Fatal(l, e)
 				return
 			}
-			log.Fatal("...deleted all offers, exiting")
+			e = fmt.Errorf("...deleted all offers, exiting")
+			logger.Fatal(l, e)
```
nikhilsaraf: can embed the error creation in the logger.Fatal line to keep things consolidated: `logger.Fatal(l, fmt.Errorf("...deleted all offers, exiting"))`

cmd/trade.go
```diff
@@ -308,49 +312,48 @@ func validateTrustlines(client *horizon.Client, botConfig *trader.BotConfig) {
...
 		for {
 			sleepSeconds := 10
-			log.Printf("sleeping for %d seconds until our deletion is confirmed and we exit...\n", sleepSeconds)
+			l.Infof("sleeping for %d seconds until our deletion is confirmed and we exit...\n", sleepSeconds)
 			time.Sleep(time.Duration(sleepSeconds) * time.Second)
 		}
 	} else {
-		log.Fatal("...nothing to delete, exiting")
+		e := fmt.Errorf("...nothing to delete, exiting")
+		logger.Fatal(l, e)
```
nikhilsaraf: consolidate into one line here too (and anywhere else where it was missed up above too)

support/logger/basicLogger.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// Error impl
+func (l *BasicLogger) Error(msg string) {
+	log.Print(msg)
+}
+
+// Efforf impl
+func (l *BasicLogger) Errorf(msg string, args ...interface{}) {
+	log.Printf(msg)
```
nikhilsaraf: also pass in the `args` here

support/logger/basicLogger.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// ensure it implements Logger
+var _ Logger = &BasicLogger{}
+
+// MakeBasicLogger is the factory method
+func MakeBasicLogger() Logger {
+	var l *BasicLogger
```
nikhilsaraf: can one-line this: `return &BasicLogger{}`

cmd/trade.go
```diff
@@ -81,9 +85,9 @@ func init() {
 
 		if *fixedIterations == 0 {
 			fixedIterations = nil
-			log.Printf("will run unbounded iterations\n")
+			l.Info("will run unbounded iterations")
```
nikhilsaraf: validateCliParams() can take in the a `l logger.Logger` as a param so it can be used here

cmd/trade.go
```diff
@@ -93,20 +97,20 @@ func init() {
 		utils.CheckConfigError(botConfig, e, *botConfigPath)
 		e = botConfig.Init()
 		if e != nil {
-			log.Println()
-			log.Fatal(e)
+			l.Info("")
```
nikhilsaraf: the first line of the `tradeCmd.Run` function should be `l := logger.MakeBasicLogger()`

this will eliminate the global variable issue

cmd/trade.go
Reidmcc: I thought the PANIC was supposed to exit, but now that you mention it I don't know why I would have thought I needed to add the exit here.

cmd/trade.go
Reidmcc: I had to put it at the beginning `func init()` because it needs to be before `validateCliParams`

Reidmcc:
Changes made

cmd/trade.go
```diff
@@ -166,10 +167,10 @@ func init() {
 		dataKey := model.MakeSortedBotKey(assetBase, assetQuote)
 		strat, e := plugins.MakeStrategy(sdex, tradingPair, &assetBase, &assetQuote, *strategy, *stratConfigPath, *simMode)
 		if e != nil {
-			log.Println()
-			log.Println(e)
+			l.Info("")
+			logger.Fatal(l, e)
```
nikhilsaraf: this shouldnot be a `logger.Fatal` because fatal means we will exit, which we don't want to do here -- see the next line where we want to delete all offers and then exit.

cmd/trade.go
```diff
@@ -81,49 +82,49 @@ func init() {
...
 		}
 
 		if *logPrefix != "" {
 			t := time.Now().Format("20060102T150405MST")
 			fileName := fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", *logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, t)
 			e = setLogFile(fileName)
 			if e != nil {
-				log.Println()
-				log.Fatal(e)
+				l.Info("")
```
nikhilsaraf: can remove the empty line printout here -- logger.Fatal starts off by printing an empty line

cmd/trade.go
```diff
@@ -192,30 +193,32 @@ func init() {
...
 		if botConfig.MonitoringPort != 0 {
 			go func() {
-				e := startMonitoringServer(botConfig)
+				e := startMonitoringServer(l, botConfig)
 				if e != nil {
-					log.Println()
-					log.Printf("unable to start the monitoring server or problem encountered while running server: %s\n", e)
+					l.Info("")
+					l.Info("unable to start the monitoring server or problem encountered while running server:")
+					logger.Fatal(l, e)
```
nikhilsaraf: here too -- logger.Fatal will cause us to exit and the next line where we delete all offers won't run

cmd/trade.go
```diff
@@ -192,30 +193,32 @@ func init() {
...
 		}
 
 		strategyFillHandlers, e := strat.GetFillHandlers()
 		if e != nil {
-			log.Println()
-			log.Printf("problem encountered while instantiating the fill tracker: %s\n", e)
-			deleteAllOffersAndExit(botConfig, client, sdex)
+			l.Info("")
+			l.Info("problem encountered while instantiating the fill tracker:")
+			logger.Fatal(l, e)
```
nikhilsaraf: here too, use Errorf instead of Fatal

cmd/trade.go
```diff
@@ -227,30 +230,31 @@ func init() {
...
-			log.Printf("Starting fill tracker with %d handlers\n", fillTracker.NumHandlers())
+			l.Infof("Starting fill tracker with %d handlers\n", fillTracker.NumHandlers())
 			go func() {
 				e := fillTracker.TrackFills()
 				if e != nil {
-					log.Println()
-					log.Printf("problem encountered while running the fill tracker: %s\n", e)
+					l.Info("")
+					l.Info("problem encountered while running the fill tracker:")
+					logger.Fatal(l, e)
```
nikhilsaraf: Errorf instead of Fatal

cmd/trade.go
```diff
@@ -227,30 +230,31 @@ func init() {
...
 					// we want to delete all the offers and exit here because we don't want the bot to run if fill tracking isn't working
-					deleteAllOffersAndExit(botConfig, client, sdex)
+					deleteAllOffersAndExit(l, botConfig, client, sdex)
 				}
 			}()
 		} else if strategyFillHandlers != nil && len(strategyFillHandlers) > 0 {
-			log.Println()
-			log.Printf("error: strategy has FillHandlers but fill tracking was disabled (set FILL_TRACKER_SLEEP_MILLIS to a non-zero value)\n")
+			l.Info("")
+			l.Info("error: strategy has FillHandlers but fill tracking was disabled (set FILL_TRACKER_SLEEP_MILLIS to a non-zero value)")
```
nikhilsaraf: Errorf instead of Info

cmd/trade.go
```diff
@@ -311,49 +314,46 @@ func validateTrustlines(client *horizon.Client, botConfig *trader.BotConfig) {
...
 	if len(dOps) > 0 {
 		e := sdex.SubmitOps(dOps, func(hash string, e error) {
 			if e != nil {
-				log.Println()
-				log.Fatal(e)
+				logger.Fatal(l, e)
 				return
 			}
-			log.Fatal("...deleted all offers, exiting")
+			e = fmt.Errorf("...deleted all offers, exiting")
```
nikhilsaraf: can embed this error in the same line as logger.Fatal

cmd/trade.go
Reidmcc: Yikes, glad you caught that.

cmd/trade.go
Reidmcc: No need for the formatting; using `Error` here

Reidmcc:
Updated

cmd/trade.go
```diff
@@ -192,30 +192,32 @@ func init() {
...
 		if botConfig.MonitoringPort != 0 {
 			go func() {
-				e := startMonitoringServer(botConfig)
+				e := startMonitoringServer(l, botConfig)
 				if e != nil {
-					log.Println()
-					log.Printf("unable to start the monitoring server or problem encountered while running server: %s\n", e)
+					l.Info("")
+					l.Info("unable to start the monitoring server or problem encountered while running server:")
+					l.Errorf("%s", e)
```
nikhilsaraf: this should typically be in the same line as where we wrap it (i.e. the line above), and should have a newline after (like we do on the left hand side of this diff). ok for now.

support/logger/logger.go
```diff
@@ -0,0 +1,25 @@
...
+	Error(msg string)
+
+	// error messages, indicates to the logger that these messages can be handled differently (different color, special format, email alerts, etc.). The type of logger will determine what to do with these messages. The logger should NOT panic on these messages. User needs to add a \n if they want a newline after the log entry.
+	Errorf(msg string, args ...interface{})
+}
+
+// Fatal is a convenience method that can be used with any Logger to log a fatal error
+func Fatal(l Logger, e error) {
+	l.Info("")
+	l.Errorf("%s", e)
```
nikhilsaraf: this should typically include a newline character at the end `l.Errorf("%s", e)` -> `l.Errorf("%s\n", e)`

nikhilsaraf:
@Reidmcc this is really good -- feel free to move forward and convert the remaining files to this new pattern.

nikhilsaraf:
this PR starts the process of addressing #9

PR #90, opened 2019-01-11
Add sdex price feed

Reidmcc:
This adds a "sdex" pricefeed option to buysell and sell, as we talked about on https://github.com/interstellar/kelp/pull/39. It doesn't have any custom strategy stuff, just the feeds.

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -164,7 +164,8 @@ func init() {
 		)
 
 		// setting the temp hack variables for the sdex price feeds
-		plugins.PrivateSdexHack = sdex
+		plugins.PrivateSdexHack.API = sdex.API
+		plugins.PrivateSdexHack.Network = sdex.Network
 
 		dataKey := model.MakeSortedBotKey(assetBase, assetQuote)
 		strat, e := plugins.MakeStrategy(sdex, tradingPair, &assetBase, &assetQuote, *strategy, *stratConfigPath, *simMode)
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -164,8 +164,13 @@ func init() {
 		)
 
 		// setting the temp hack variables for the sdex price feeds
-		plugins.PrivateSdexHack.API = client
-		plugins.PrivateSdexHack.Network = utils.ParseNetwork(botConfig.HorizonURL)
+		e = plugins.SetPrivateSdexHack(client, utils.ParseNetwork(botConfig.HorizonURL))
+		if e != nil {
+			l.Info("")
+			l.Errorf("%s", e)
+			// we want to delete all the offers and exit here since there is something wrong with our setup
+			deleteAllOffersAndExit(l, botConfig, client, sdex)
+		}
 
 		dataKey := model.MakeSortedBotKey(assetBase, assetQuote)
 		strat, e := plugins.MakeStrategy(sdex, tradingPair, &assetBase, &assetQuote, *strategy, *stratConfigPath, *simMode)
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -164,8 +164,8 @@ func init() {
 		)
 
 		// setting the temp hack variables for the sdex price feeds
-		plugins.PrivateSdexHack.API = sdex.API
-		plugins.PrivateSdexHack.Network = sdex.Network
+		plugins.PrivateSdexHack.API = client
+		plugins.PrivateSdexHack.Network = utils.ParseNetwork(botConfig.HorizonURL)
 
 		dataKey := model.MakeSortedBotKey(assetBase, assetQuote)
 		strat, e := plugins.MakeStrategy(sdex, tradingPair, &assetBase, &assetQuote, *strategy, *stratConfigPath, *simMode)
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -163,6 +163,9 @@ func init() {
 			sdexAssetMap,
 		)
 
+		// setting the temp hack variables for the sdex price feeds
+		plugins.PrivateSdexHack = sdex
+
 		dataKey := model.MakeSortedBotKey(assetBase, assetQuote)
 		strat, e := plugins.MakeStrategy(sdex, tradingPair, &assetBase, &assetQuote, *strategy, *stratConfigPath, *simMode)
 		if e != nil {
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/interstellar/kelp/model"
 	"github.com/interstellar/kelp/plugins"
+	"github.com/interstellar/kelp/support/logger"
 	"github.com/interstellar/kelp/support/monitoring"
 	"github.com/interstellar/kelp/support/networking"
 	"github.com/interstellar/kelp/support/utils"
@@ -45,10 +46,10 @@ func hiddenFlag(flag string) {
 	}
 }
 
-func logPanic() {
+func logPanic(l logger.Logger) {
 	if r := recover(); r != nil {
 		st := debug.Stack()
-		log.Printf("PANIC!! recovered to log it in the file\npanic: %v\n\n%s\n", r, string(st))
+		l.Errorf("PANIC!! recovered to log it in the file\npanic: %v\n\n%s\n", r, string(st))
 	}
 }
 
@@ -70,7 +71,7 @@ func init() {
 	hiddenFlag("operationalBufferNonNativePct")
 	tradeCmd.Flags().SortFlags = false
 
-	validateCliParams := func() {
+	validateCliParams := func(l logger.Logger) {
 		if *operationalBuffer < 0 {
 			panic(fmt.Sprintf("invalid operationalBuffer argument, must be non-negative: %f", *operationalBuffer))
 		}
@@ -81,20 +82,20 @@ func init() {
 
 		if *fixedIterations == 0 {
 			fixedIterations = nil
-			log.Printf("will run unbounded iterations\n")
+			l.Info("will run unbounded iterations")
 		} else {
-			log.Printf("will run only %d update iterations\n", *fixedIterations)
+			l.Infof("will run only %d update iterations\n", *fixedIterations)
 		}
 	}
 
 	tradeCmd.Run = func(ccmd *cobra.Command, args []string) {
+		l := logger.MakeBasicLogger()
 		var botConfig trader.BotConfig
 		e := config.Read(*botConfigPath, &botConfig)
 		utils.CheckConfigError(botConfig, e, *botConfigPath)
 		e = botConfig.Init()
 		if e != nil {
-			log.Println()
-			log.Fatal(e)
+			logger.Fatal(l, e)
 		}
 
 		if *logPrefix != "" {
@@ -102,28 +103,27 @@ func init() {
 			fileName := fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", *logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, t)
 			e = setLogFile(fileName)
 			if e != nil {
-				log.Println()
-				log.Fatal(e)
+				logger.Fatal(l, e)
 				return
 			}
-			log.Printf("logging to file: %s", fileName)
+			l.Infof("logging to file: %s\n", fileName)
 
 			// we want to create a deferred recovery function here that will log panics to the log file and then exit
-			defer logPanic()
+			defer logPanic(l)
 		}
 
 		startupMessage := "Starting Kelp Trader: " + version + " [" + gitHash + "]"
 		if *simMode {
 			startupMessage += " (simulation mode)"
 		}
-		log.Println(startupMessage)
+		l.Info(startupMessage)
 
 		// now that we've got the basic messages logged, validate the cli params
-		validateCliParams()
+		validateCliParams(l)
 
 		// only log botConfig file here so it can be included in the log file
 		utils.LogConfig(botConfig)
-		log.Printf("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
+		l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
 		client := &horizon.Client{
 			URL:  botConfig.HorizonURL,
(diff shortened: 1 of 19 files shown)


plugins/priceFeed.go
```diff
@@ -38,18 +38,24 @@ func MakePriceFeed(feedType string, url string) (api.PriceFeed, error) {
...
+		}
+		return SDEXfeed, nil
 	}
 	return nil, nil
 }
 
 // MakeFeedPair is the factory method that we expose
-func MakeFeedPair(dataTypeA, dataFeedAUrl, dataTypeB, dataFeedBUrl string) (*api.FeedPair, error) {
-	feedA, e := MakePriceFeed(dataTypeA, dataFeedAUrl)
+func MakeFeedPair(sdex *SDEX, dataTypeA, dataFeedAUrl, dataTypeB, dataFeedBUrl string) (*api.FeedPair, error) {
```
nikhilsaraf: we should not take in `sdex` here.

The `priceFeed.go` should already hold a reference to the sdex instance before we ever would want to call MakePriceFeed or MakeFeedPair. However, we're not currently set up to allow for that so you can just keep a global variable in priceFeed.go for now, calling it `_privateSdexHack` which you can set in `trade.go` after constructing sdex.go. Please add a comment on the line where you set it to indicate this is a temporary hack.

The right way to do this would be to make an `objectFactory.go` struct that is stateful, containing the sdex instance and all factory methods should hang off this instance (i.e. all Make* methods)

plugins/sellStrategy.go
```diff
@@ -39,6 +39,7 @@ func makeSellStrategy(
 	config *sellConfig,
 ) (api.Strategy, error) {
 	pf, e := MakeFeedPair(
+		sdex,
```
nikhilsaraf: all these changes would be reverted once we add the `sdex` instance to `priceFeed.go`

plugins/sdexFeed.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// GetPrice returns the SDEX mid price for the trading pair
+func (s *sdexFeed) GetPrice() (float64, error) {
+	orderBook, e := GetOrderBook(s.sdex.API, s.assetBase, s.assetQuote)
+	if e != nil {
+		return 0, fmt.Errorf("unable to get sdex price: %s", e)
+	}
+	bids := orderBook.Bids
+	topBidPrice := utils.PriceAsFloat(bids[0].Price)
```
nikhilsaraf: can you add empty lines here: after error check, after topBidPrice, after lowAskPrice?

plugins/cmcFeed.go
```diff
@@ -45,8 +45,8 @@ type cmcFeed struct {
 // ensure that it implements PriceFeed
 var _ api.PriceFeed = &cmcFeed{}
 
-// newCMCFeed creates a new CMC Feed from a URL
-func newCMCFeed(url string) *cmcFeed {
+// NewCMCFeed creates a new CMC Feed from a URL
+func NewCMCFeed(url string) *cmcFeed {
```
nikhilsaraf: this would be reverted

plugins/sdex.go
```diff
@@ -813,3 +813,13 @@ func (sdex *SDEX) GetLatestTradeCursor() (interface{}, error) {
 
 	return records[0].PT, nil
 }
+
+// GetOrderBook gets the SDEX order book
+func GetOrderBook(api *horizon.Client, assetBase *horizon.Asset, assetQuote *horizon.Asset) (orderBook horizon.OrderBookSummary, e error) {
```
nikhilsaraf: This should implement the method signature from `TradeAPI` in `exchange.go` as closely as possible:

    GetOrderBook(pair *model.TradingPair, maxCount int32) (*model.OrderBook, error)

sdex.go holds a `tradingPair` object, we should just use that when creating the OrderBook and any related Orders. For now, we should also return an error if the passed in `pair` is different than the `tradingPair` saved.

---

also as a general kelp convention we can leave the return values in the signatures unnamed -- the convention is to only name them if they are of the same type so it's a better API. here they are different types so we can leave them unnamed

Reidmcc: I wrote it this way specifically to allow for non-matching feeds; do you want to disallow using another asset's feed for now?

nikhilsaraf: ah, I see. that makes sense.

In that case I'd recommend creating a new sdex instance in the `case` statement of priceFeed.go (no global var needed as commented above). Here's a list of the all the inputs to the factory method, we can stub out those values that are not needed for this call:

    func MakeSDEX(
        api *horizon.Client,    // this can be a global var for now _sdexHack.api
        sourceSeed string,    // can be ""
        tradingSeed string,    // can be ""
        sourceAccount string,    // can be ""
        tradingAccount string,    // can be ""
        network build.Network,    // this can be a global var for now _sdexHack.network
        threadTracker *multithreading.ThreadTracker,    // this can be nil
        operationalBuffer float64,    // can be 0
        operationalBufferNonNativePct float64,    // can be 0
        simMode bool,    // can be true
        pair *model.TradingPair,    // should be a valid pair, see trade.go for how we make this
        assetMap map[model.Asset]horizon.Asset,    // should be a valid map for the two assets, see trade.go for how we construct this
    )

plugins/sdexFeed.go
```diff
@@ -0,0 +1,79 @@
...
+	s := new(sdexFeed)
+	s.sdex = sdex
+	urlParts := strings.Split(url, "/")
+
+	baseURL := strings.Split(urlParts[0], ":")
+	baseCode := baseURL[0]
+	baseIssuer := baseURL[1]
+	baseConvert, e := parseAsset(baseCode, baseIssuer)
+	if e != nil {
+		return nil, fmt.Errorf("unable to convert base asset url to sdex asset")
```
nikhilsaraf: we should never _hide_ errors, this should display the error: 

    return nil, fmt.Errorf("unable to convert base asset url to sdex asset: %s", e)

plugins/priceFeed.go
nikhilsaraf: see new comment on instantiating sdex directly here instead of using the global variable approach -- we would still need a temporary global variable for certain parts (network and horizon api) to allow instantiating the sdex object

plugins/sdexFeed.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	bids := orderBook.Bids
+	topBidPrice := utils.PriceAsFloat(bids[0].Price)
+	asks := orderBook.Asks
+	lowAskPrice := utils.PriceAsFloat(asks[0].Price)
+	centerPrice := (topBidPrice + lowAskPrice) / 2
+	return centerPrice, nil
+}
+
+func parseAsset(code string, issuer string) (*horizon.Asset, error) {
```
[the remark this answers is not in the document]
nikhilsaraf: ok, makes sense to generalize -- then the logic in trader/config.go#Init() should wrap the returned error from parseAsset() with the related "A" or "B" error message instead of just returning the error (trader/config.go line 81 and line 87)

plugins/sdexFeed.go
Reidmcc: Agreed, oversight on my part

plugins/priceFeed.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/interstellar/kelp/model"
 )
 
+// PrivateSdexHack is a temporary hack variable for SDEX price feeds pending refactor
+var PrivateSdexHack *SDEX
```
nikhilsaraf: with the comment below on sdex.go#GetOrderBook, we should be constructing the SDEX instance directly in the case statement below. i.e. we no longer need the global variable for SDEX but will instead need a global variable for the dependencies needed to instantiate a new SDEX instance:
- api *horizon.Client
- network build.Network

you can keep all this in one global variable struct if you want, calling it PrivateSdexHack like is already done.

    var PrivateSdexHack struct {
        api *horizon.Client
        network build.Network
    }

This will ensure that your logic in newSDEXFeed calls the `api` and `network` values on PrivateSdexHack (it probably throws a compile error now?)

plugins/sdex.go
nikhilsaraf: we should be able to change this function signature now to:

    GetOrderBook(pair *model.TradingPair, maxCount int32) (*model.OrderBook, error)

where we can use the tradingPair in sdex.go where needed

plugins/sdexFeed.go
```diff
@@ -0,0 +1,90 @@
...
+		tradingPair.Quote: *quoteConvert,
+	}
+
+	feedSDEX := MakeSDEX(
+		PrivateSdexHack.API,
+		"",
+		"",
+		"",
+		"",
+		PrivateSdexHack.Network,
```
nikhilsaraf: this looks great!

plugins/sdexFeed.go
```diff
@@ -0,0 +1,90 @@
...
+		Base:  model.Asset(utils.Asset2CodeString(*baseConvert)),
+		Quote: model.Asset(utils.Asset2CodeString(*quoteConvert)),
+	}
+
+	sdexAssetMap := map[model.Asset]horizon.Asset{
+		tradingPair.Base:  *baseConvert,
+		tradingPair.Quote: *quoteConvert,
+	}
+
+	feedSDEX := MakeSDEX(
```
nikhilsaraf: can call this `sdex` instead of `feedSDEX` to keep the names as terse as needed

trader/config.go
```diff
@@ -75,15 +74,15 @@ func (b *BotConfig) Init() error {
 		return fmt.Errorf("error: both assets cannot be the same '%s:%s'", b.AssetCodeA, b.IssuerA)
 	}
 
-	asset, e := parseAsset(b.AssetCodeA, b.IssuerA, "A")
+	asset, e := utils.ParseAsset(b.AssetCodeA, b.IssuerA)
 	if e != nil {
-		return e
+		return fmt.Errorf("Error while parsing Asset A: %s", e)
```
nikhilsaraf: :+1:

plugins/priceFeed.go
Reidmcc: Nope, it compiles and runs. I do have it explicitly grabbing  `.network` and `.api` so that may be why it works.

nikhilsaraf:
@Reidmcc let me know once you've had a chance to address the two points in the comment above.
Created a github issue to track and reference this: https://github.com/interstellar/kelp/issues/97

nikhilsaraf:
closes #97

plugins/priceFeed.go
Reidmcc: Change made

plugins/sdexFeed.go
Reidmcc: done

plugins/priceFeed.go
```diff
@@ -38,6 +41,12 @@ func MakePriceFeed(feedType string, url string) (api.PriceFeed, error) {
 		}
 		tickerAPI := api.TickerAPI(exchange)
 		return newExchangeFeed(url, &tickerAPI, &tradingPair), nil
+	case "sdex":
+		sdexFeed, e := newSDEXFeed(url)
```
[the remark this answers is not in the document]
Reidmcc: Done. Should we also change the other price feed functions from `new` to `make`?

plugins/sdex.go
Reidmcc: done

plugins/priceFeed.go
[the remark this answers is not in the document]
nikhilsaraf: not at this point since it's not a big deal right now. can address later if needed as a separate issue.

cmd/trade.go
```diff
@@ -163,6 +163,10 @@ func init() {
 			sdexAssetMap,
 		)
 
+		// setting the temp hack variables for the sdex price feeds
+		plugins.PrivateSdexHack.API = sdex.API
```
nikhilsaraf: this should be `client` (not `sdex.API`)  and `utils.ParseNetwork(botConfig.HorizonURL)` instead of `sdex.Network` (can extract into a `network` variable first.

it's reasonable for MakeSDEX to do something like this if it needs to: `API = nil` since that's an abstraction we get with the make factory method. So when we reach into the sdex.API here we're breaking that abstraction and creating a dependency where there shouldn't be.

plugins/priceFeed.go
```diff
@@ -6,8 +6,16 @@ import (
 
 	"github.com/interstellar/kelp/api"
 	"github.com/interstellar/kelp/model"
+	"github.com/stellar/go/build"
+	"github.com/stellar/go/clients/horizon"
 )
 
+// PrivateSdexHack is a temporary hack variable for SDEX price feeds pending refactor
+var PrivateSdexHack struct {
```
nikhilsaraf: :+1:

plugins/sdex.go
```diff
@@ -813,3 +813,19 @@ func (sdex *SDEX) GetLatestTradeCursor() (interface{}, error) {
 
 	return records[0].PT, nil
 }
+
+// GetOrderBook gets the SDEX order book
+func (sdex *SDEX) GetOrderBook(pair *model.TradingPair) (orderBook horizon.OrderBookSummary, e error) {
+	//b, e := api.LoadOrderBook(*assetBase, *assetQuote)
```
nikhilsaraf: can remove commented out code for now

plugins/sdex.go
```diff
@@ -813,3 +813,19 @@ func (sdex *SDEX) GetLatestTradeCursor() (interface{}, error) {
 
 	return records[0].PT, nil
 }
+
+// GetOrderBook gets the SDEX order book
+func (sdex *SDEX) GetOrderBook(pair *model.TradingPair) (orderBook horizon.OrderBookSummary, e error) {
```
nikhilsaraf: this should conform to the API method described in the previous comments:

```GetOrderBook(pair *model.TradingPair, maxCount int32) (*model.OrderBook, error)```

this would require transforming orders from horizon's format to the `model.OrderBook` format
I can give you sample code for this if needed, I have something.

Comment (nikhilsaraf) [reply plugins/sdex.go]: @Reidmcc the sample code below converts a single `manageOffer` to a `model.Order`.
You'd need to do something similar, converting from `horizon.OrderBookSummary` to `model.OrderBook` (which mostly contains a list of `model.Order`). Hope this is useful.

```golang
func manageOffer2Order(mob *build.ManageOfferBuilder, baseAsset horizon.Asset, quoteAsset horizon.Asset) (*model.Order, error) {
	orderAction := model.OrderActionSell
	price := model.NumberFromFloat(float64(mob.MO.Price.N)/float64(mob.MO.Price.D), utils.SdexPrecision)
	volume := model.NumberFromFloat(float64(mob.MO.Amount)/math.Pow(10, 7), utils.SdexPrecision)
	isBuy, e := assetsEqual(quoteAsset, mob.MO.Selling)
	if e != nil {
		return nil, fmt.Errorf("could not compare assets, error: %s", e)
	}
	if isBuy {
		orderAction = model.OrderActionBuy
		// TODO need to test price and volume conversions correctly
		// volume calculation needs to happen first since it uses the non-inverted price when multiplying
		volume = model.NumberFromFloat(volume.AsFloat()*price.AsFloat(), utils.SdexPrecision)
		price = model.InvertNumber(price)
	}

	return &model.Order{
		Pair: &model.TradingPair{
			Base:  model.FromHorizonAsset(baseAsset),
			Quote: model.FromHorizonAsset(quoteAsset),
		},
		OrderAction: orderAction,
		OrderType:   model.OrderTypeLimit,
		Price:       price,
		Volume:      volume,
		Timestamp:   model.MakeTimestamp(time.Now().UnixNano() / int64(time.Millisecond)),
	}, nil
}
```

PR #91, opened 2019-01-13
add CCXT's FetchOHLCV call

Reidmcc:
I'm thinking about indicator-based strategies, and CCXT has a fetchOHLCV call for candlestick data that wasn't in our _support/sdk/cctx.go_. This supports the historical data feature https://github.com/interstellar/kelp/issues/10.

Of our supported CCXT exchanges, it works for Binance and Bittrex; Poloniex doesn't recognize this CCXT call.

### DIFF
--- a/support/sdk/ccxt.go
+++ b/support/sdk/ccxt.go
@@ -301,6 +301,7 @@ func (c *Ccxt) FetchTrades(tradingPair string) ([]CcxtTrade, error) {
 	return output, nil
 }
 
+// FetchMyTrades calls the /fetchMyTrades endpoint on CCXT
 func (c *Ccxt) FetchMyTrades(tradingPair string) ([]CcxtTrade, error) {
 	e := c.symbolExists(tradingPair)
 	if e != nil {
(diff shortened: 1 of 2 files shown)


nikhilsaraf:
closes #98

support/sdk/ccxt.go
```diff
@@ -514,3 +515,56 @@ func (c *Ccxt) CancelOrder(orderID string, tradingPair string) (*CcxtOpenOrder,
...
+	if e != nil {
+		return nil, fmt.Errorf("error fetching candlestick data: %s", e)
+	}
+
+	outputList := output.([]interface{})
+	var candleList []CcxtCandle
+
+	for _, c := range outputList {
+		rawOCHLCV := c.([]interface{})
+		var candle CcxtCandle
```
nikhilsaraf: always best to assign vars in a single line: `candle := CcxtCandle ...`

support/sdk/ccxt.go
```diff
@@ -514,3 +515,56 @@ func (c *Ccxt) CancelOrder(orderID string, tradingPair string) (*CcxtOpenOrder,
...
+		candle = CcxtCandle{
+			TimeStamp: int64(rawOCHLCV[0].(float64)),
```
nikhilsaraf: do we need any error checks here in case certain values are missing?
what if rawOHLCV has less than 6 items in the list?

support/sdk/ccxt.go
```diff
@@ -514,3 +515,56 @@ func (c *Ccxt) CancelOrder(orderID string, tradingPair string) (*CcxtOpenOrder,
...
+	e = networking.JSONRequest(c.httpClient, "POST", url, string(data), map[string]string{}, &output)
+	if e != nil {
+		return nil, fmt.Errorf("error fetching candlestick data: %s", e)
+	}
+
+	outputList := output.([]interface{})
+	var candleList []CcxtCandle
+
+	for _, c := range outputList {
+		rawOCHLCV := c.([]interface{})
```
nikhilsaraf: rawOCHLCV -> rawOHLCV

support/sdk/ccxt.go
```diff
@@ -514,3 +515,56 @@ func (c *Ccxt) CancelOrder(orderID string, tradingPair string) (*CcxtOpenOrder,
...
+	url := c.ccxtBaseURL + pathExchanges + "/" + c.exchangeName + "/" + c.instanceName + "/fetchOHLCV"
+	// decode generic data (see "https://blog.golang.org/json-and-go#TOC_4.")
+	var output interface{}
+	e = networking.JSONRequest(c.httpClient, "POST", url, string(data), map[string]string{}, &output)
+	if e != nil {
+		return nil, fmt.Errorf("error fetching candlestick data: %s", e)
+	}
+
+	outputList := output.([]interface{})
+	var candleList []CcxtCandle
```
nikhilsaraf: can put can outputList's type assertion inlined into the range. Can keep candleList's declaration closer to the for loop (i.e. remove newline after candleList), since we're using the for loop to build the contents of the candleList so they should be together rather than separated by a space.

support/sdk/ccxt_test.go
```diff
@@ -647,3 +647,31 @@ func TestCancelOrder(t *testing.T) {
...
+		return
+	}
+
+	m, e := c.FetchOHLCV("BTC/USDT", "1h")
+	if e != nil {
+		assert.Fail(t, fmt.Sprintf("error when fetching tickers: %s", e))
+		return
+	}
+
+	assert.IsType(t, []CcxtCandle{}, m)
```
nikhilsaraf: we don't need this type assertion

support/sdk/ccxt.go
Reidmcc: if ccxt returned data without throwing an error it seems unlikely that there would be the wrong number of objects; can't hurt to be safe though

Reidmcc:
@nikhilsaraf done, though I couldn't add it to ccxtExchange_test.go because that only wants to test things that api.Exchange has, and I don't think we want to add this to api.Exchange since not every exchange necessarily has OHLCV data; the SDEX doesn't

PR #100, opened 2019-02-02
Add reworked stop-limit strategy

Reidmcc:
This pr is the reworked replacement for https://github.com/interstellar/kelp/pull/84, which was an implementation compatible with older Kelp code. This pr implements the stop-limit functionality for https://github.com/interstellar/kelp/issues/83.

### DIFF
--- a/plugins/stoplimitLevelProvider.go
+++ b/plugins/stoplimitLevelProvider.go
@@ -1,0 +1,152 @@
+package plugins
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/interstellar/kelp/api"
+	"github.com/interstellar/kelp/model"
+	"github.com/interstellar/kelp/support/utils"
+	"github.com/stellar/go/clients/horizon"
+)
+
+// stopLimitLevelProvider generates a level if the stop condition is met
+type stopLimitLevelProvider struct {
+	pf               *api.FeedPair
+	assetBase        *horizon.Asset
+	assetQuote       *horizon.Asset
+	amountOfBase     float64
+	stopPrice        float64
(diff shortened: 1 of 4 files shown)


plugins/stoplimitLevelProvider.go
```diff
@@ -0,0 +1,152 @@
...
+
+// getFirstOffer gets the trading pair's top bid or low ask
+func (p *stopLimitLevelProvider) getFirstOffer(side string) (float64, error) {
+	var topPrice float64
+	orderBook, e := p.sdex.GetOrderBook(p.sdex.pair, 1)
+	if e != nil {
+		return 0, fmt.Errorf("unable to get sdex price: %s", e)
+	}
+
+	if side == "bid" {
```
nikhilsaraf: should use OrderActionBuy and OrderActionSell for bid/ask here

plugins/stoplimitStrategy.go
```diff
@@ -0,0 +1,75 @@
...
+	)
+	if e != nil {
+		return nil, fmt.Errorf("cannot make the stoplimit strategy because we could not make the feed pair: %s", e)
+	}
+	sellSideStrategy := makeSellSideStrategy(
+		sdex,
+		orderConstraints,
+		assetBase,
+		assetQuote,
+
```
nikhilsaraf: extra newline here

we should check that the limit price is below the stop price

nikhilsaraf:
closing this PR for now as we had discussed @Reidmcc .
This would need to be able to allow a stream of stop limit orders. Will be more useful once we have the GUI up and running so it can be more useful.

Issue #109, opened 2019-02-20
ccxt-rest has correct prices, yet kelp does not


paul-token:
## Describe the bug

When using ccxt via `DATA_FEED_B_URL="ccxt-poloniex/XRP/USDT"` in buysell.cfg, kelp's logs are showing ```bidPrice=0.0000000, askPrice=0.0000000, centerPrice=0.0000000```, whereas querying ccxt directly results in correct prices. 

This is how I queried ccxt-rest directly:

```
curl -X POST http://localhost:3000/exchanges/binance/binance/fetchTicker -d '["XRP/USDT"]'
```

## Expected behavior

kelp should pull in the right prices

## Frequency

Seems to be always

## Steps To Reproduce

Set up ccxt via `DATA_FEED_B_URL="ccxt-poloniex/XRP/USDT"` in a buysell.cfg

## Your Environment

  version: master:v1.3.0
  git branch: master
  git hash: e19784fc9cdb113d5b51c8f421c8ab2d49752dc0
  build date: 20190111T004556Z
  GOOS: linux
  GOARCH: amd64

## Context

We are unable to use kelp for trading.

paul-token:
Update: I just noticed this bug happen for a different trading pair. All I did was restart kelp (to change the `DATA_FEED_B_URL`), and then the same `bidPrice=0.0000000, askPrice=0.0000000, centerPrice=0.0000000` issue appeared.

Here are the logs of when things went from normal to busted:

```
Feb 21 03:07:52 steve-kelp bash[13536]: 2019/02/21 03:07:52 submitting tx XDR to network (async)
Feb 21 03:07:55 steve-kelp bash[13536]: 2019/02/21 03:07:55 (async) tx confirmation hash: 12b27bb14e6e3dc6eb051e875c64c3654f5bc55abfd74a4625b3c470b6b54a6a
Feb 21 03:07:55 steve-kelp bash[13536]: 2019/02/21 03:07:55 ----------------------------------------------------------------------------------------------------
Feb 21 03:07:55 steve-kelp bash[13536]: 2019/02/21 03:07:55 sleeping for 3m43.020791558s...
Feb 21 03:11:38 steve-kelp bash[13536]: 2019/02/21 03:11:38 intervalTimeController tickInterval=3m0s, shouldUpdate=true, elapsedSinceUpdate=3m50.962134257s
[...]

Issue #110, opened 2019-02-21
kelp showing/using overly rounded/low-preicision prices


paul-token:
## Describe the bug

When using ccxt via `DATA_FEED_B_URL="ccxt-poloniex/ETH/USDT"` in buysell.cfg, kelp's logs are showing ```2019/02/21 00:56:42 price from exchange feed (ccxt-bittrex/ETH/USDT): bidPrice=147.0000000, askPrice=148.0000000, centerPrice=147.5000000```, which seems to be overly rounded, whereas querying ccxt directly results in correct/more-precise prices. 

## Expected behavior

kelp should pull in (and use) more accurate prices when possible (for example, when using sdex)

## Frequency

Seems to be always

## Steps To Reproduce

Set up ccxt via `DATA_FEED_B_URL="ccxt-poloniex/ETH/USDT"` in a buysell.cfg

## Your Environment

  version: fixed_fee:v1.4.0-3-ga0362947
  git branch: fixed_fee
  git hash: a036294782df939156167cb65c564aab3f62fbdd
  build date: 20190220T231403Z
  GOOS: linux
  GOARCH: amd64

## Context

With too low a precision value there are many cases where no trades will ever be made even though there could be a ~1% difference in certain assets (for example, with ETH)

PR #112, opened 2019-02-22
load orderConstraints for ccxtExchange from CCXT API

nikhilsaraf:


nikhilsaraf:
closes https://github.com/stellar/kelp/issues/109 and https://github.com/stellar/kelp/issues/110
fyi @paul-token @skirsch

Issue #139, opened 2019-03-31
accomodate additional ccxt parameters


Reidmcc:
## Desired Behavior

Some ccxt exchanges require additional parameters in API requests; e.g. Coinbase Pro requires a passphrase in an http header as well as an API key and secret. Currently Kelp doesn't have a way to get those parameters through ccxt and into the API requests.

## Impact

Without this capability Kelp can't trade on Coinbase Pro, or any other exchanges that require additional parameters.

## Feature Suggestion

We can implement this by adding a place to specify additional ccxt parameters in the trader configuration file to pass to the local ccxt server and add code to Kelp to pass them. Pull request incoming.

## Additional context

I originally tried to accomplish this by adding capability to append http headers in support/sdk/ccxt.go. This didn't work because it turns out the error was coming from the ccxt local server; the request never went to Coinbase. Ccxt knows about the Coinbase passphrase and requires it before it will send requests. The correct method was to pass parameters to ccxt in the format it requires. 

I ended up leaving the header appending feature in because it will allow users to add any headers that ccxt *doesn't* know about.

nikhilsaraf:
closed by https://github.com/stellar/kelp/pull/140

PR #140, opened 2019-03-31
Enable additional ccxt params and http headers

Reidmcc:
This pr implements the feature requested in https://github.com/stellar/kelp/issues/139 by allowing users to specify any additional ccxt parameters and/or http headers they may need to add to their API requests. I have live tested it on Coinbase Pro, so if this is implemented we can marked Coinbase Pro as tested.

### DIFF
--- a/api/exchange.go
+++ b/api/exchange.go
@@ -73,6 +73,8 @@ type FillTrackable interface {
 type Constrainable interface {
 	// return nil if the constraint does not exist for the exchange
 	GetOrderConstraints(pair *model.TradingPair) *model.OrderConstraints
+
+	OverrideOrderConstraints(pair *model.TradingPair, override *model.OrderConstraintsOverride)
 }
 
 // OrderbookFetcher extracts out the method that should go into ExchangeShim for now
--- a/api/exchange.go
+++ b/api/exchange.go
@@ -16,8 +16,8 @@ type ExchangeAPIKey struct {
 
 // ExchangeParam specifies an additional parameter to be sent when initializing the exchange
 type ExchangeParam struct {
-	Parameter string
-	Value     string
+	Param string
+	Value string
 }
 
 // ExchangeHeader specifies additional HTTP headers
--- a/api/exchange.go
+++ b/api/exchange.go
@@ -14,6 +14,18 @@ type ExchangeAPIKey struct {
 	Secret string
 }
 
+// ExchangeParam specifies an additional parameter to be sent when initializing the exchange
+type ExchangeParam struct {
+	Parameter string
+	Value     string
+}
+
+// ExchangeHeader specifies additional HTTP headers
+type ExchangeHeader struct {
+	Header string
+	Value  string
+}
+
 // Account allows you to access key account functions
 type Account interface {
 	GetAccountBalances(assetList []interface{}) (map[interface{}]model.Number, error)
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -93,8 +93,19 @@ func validateBotConfig(l logger.Logger, botConfig trader.BotConfig) {
 		logger.Fatal(l, fmt.Errorf("The `FEE` object needs to exist in the trader config file when trading on SDEX"))
 	}
 
-	if !botConfig.IsTradingSdex() && botConfig.MinCentralizedBaseVolume == 0.0 {
-		logger.Fatal(l, fmt.Errorf("need to specify non-zero MIN_CENTRALIZED_BASE_VOLUME config param in trader config file when not trading on SDEX"))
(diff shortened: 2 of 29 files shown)


api/exchange.go
```diff
@@ -14,6 +14,18 @@ type ExchangeAPIKey struct {
 	Secret string
 }
 
+// CcxtParam specifies an additional ccxt parameter
+type CcxtParam struct {
```
nikhilsaraf: @Reidmcc mostly looks good. can you rename `CcxtParam` to `ExchangeParam` everywhere?
The exchange interface abstracts away the idea of ccxt / sdex / kraken from the calling code.

api/exchange.go
```diff
@@ -14,6 +14,18 @@ type ExchangeAPIKey struct {
...
+	Parameter string
+	Value     string
+}
+
+// ExchangeHeader specifies additional HTTP headers for centralized exchanges
```
nikhilsaraf: same logic as comment above, this comment should not say "centralized exchanges" since it abstracts away the concept of the underlying exchange.

cmd/trade.go
```diff
@@ -187,8 +187,24 @@ func makeExchangeShimSdex(
 			})
 		}
 
+		ccxtParams := []api.CcxtParam{}
```
nikhilsaraf: similarly, this here should be `exchangeParams` instead of `ccxtParams`

plugins/ccxtExchange.go
```diff
@@ -24,6 +24,7 @@ type ccxtExchange struct {
 	delimiter        string
 	orderConstraints map[model.TradingPair]model.OrderConstraints
 	api              *sdk.Ccxt
+	headers          []api.ExchangeHeader
```
nikhilsaraf: @Reidmcc  is `headers` here unused?

plugins/ccxtExchange_test.go
```diff
@@ -14,6 +14,8 @@ import (
 
 var supportedExchanges = []string{"binance"}
 var emptyAPIKey = api.ExchangeAPIKey{}
+var emptyParams = api.CcxtParam{}
+var emptyHeader = api.ExchangeHeader{}
```
nikhilsaraf: @Reidmcc  there are a few tests here (`ccxtExchange_test.go`) that include the list of "tested" exchanges (example: `TestGetOrderConstraints_Ccxt_Precision`). Can you add coinbasepro to that test?

plugins/ccxtExchange.go
Reidmcc: yep, leftover from an earlier draft

Reidmcc:
@nikhilsaraf changes made

Reidmcc:
ok, apparently I need to go over the `ccxtExchange_test.go` changes more carefully. I'll try to do that tonight.

plugins/ccxtExchange_test.go
```diff
@@ -471,11 +473,17 @@ func TestGetOrderConstraints_Ccxt_Precision(t *testing.T) {
 			wantPricePrecision: 8,
 			wantVolPrecision:   0,
 		},
+		{
+			exchangeName:       "coinbasepro",
+			pair:               &model.TradingPair{Base: model.XLM, Quote: model.USD},
+			wantPricePrecision: 6,
+			wantVolPrecision:   0,
```
nikhilsaraf: can you share the link where to find precision values for coinbasepro here? may be useful to put this as a comment inside each of these test cases
(don't worry about the kraken/binance cases, I'll add these links in)

Reidmcc:
@nikhilsaraf alright, it's fixed. It was failing because of the coinbase pro precision problems

nikhilsaraf:
@Reidmcc I've made some changes to merge the master branch in and fix some remaining comments and convention. I think this is pretty much good to go, could you test it out once? -- the precision override logic should also be included here so it should all just work.

Also, I could not find the coinbase pro API docs that show the headers and params that are needed. Do you have the link handy so I can add it in?

Reidmcc:
@nikhilsaraf the parameters pass in fine, but the volume override isn't working correctly. Kelp still calculates volumes to SDEX precision, and when inverting for buy orders sometimes ends up with low over-precise values that can fail min volume tests if you set your volume to the minimum, i.e. if you set your volume to 1.0, Kelp may invert the volume to 0.99999981, and then reject the order at `orderConstraintsFilter`.

[mtest_XLM_USD_20190413T134346CDT.log](https://github.com/stellar/kelp/files/3076624/mtest_XLM_USD_20190413T134346CDT.log)

Reidmcc:
@nikhilsaraf cbpro header reqs are here https://docs.pro.coinbase.com/#creating-a-request

nikhilsaraf:
@Reidmcc filed an issue for the incorrect volume amounts here: https://github.com/stellar/kelp/issues/155

For now I'd recommend using an `AMOUNT_OF_A_BASE` value that is slightly higher than the `CENTRALIZED_MIN_BASE_VOLUME_OVERRIDE` amount

PR #168, opened 2019-05-06
Build and test Kelp with Semaphore CI

accordeiro:
This PR aims to create an initial config for building and testing Kelp with Semaphore CI.

We're thinking about switching to Semaphore because it runs multiple builds in parallel, and it would greatly improve our workflow, especially in projects with many contributors such as the [Go monorepo](https://github.com/stellar/go).

**Note:** Once this PR gets merged, we'll need to do some setup on Semaphore to make sure it starts building stellar/kelp.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -1,0 +1,56 @@
+version: v1.0
+name: First pipeline example
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+blocks:
+  - name: "Build"
+    task:
+      env_vars:
+        - name: APP_ENV
+          value: prod
+      jobs:
+      - name: Docker build
+        commands:
+          - checkout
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -0,0 +1,64 @@
...
+          - glide install
+
+          # Installing node.js 10.5:
+          - nvm install 10.5 && nvm use 10.5
+
+          # Running CCXT:
+          - docker run -p 3000:3000 -d franzsee/ccxt-rest:v0.0.4
+      jobs:
+        # Run the test suite + try building Kelp with Go 1.10
+        - name: Run package tests (1.10)
```
nikhilsaraf: I'm guessing that there's no way to avoid this repetition for testing on multiple versions of golang?
Do you know if we can have templates or something to avoid this repetition?

Issue #156, opened 2019-04-15
[3] support dynamic CCXT headers for exchanges such as Coinbase


nikhilsaraf:
coinbasepro header reqs are here: https://docs.pro.coinbase.com/#creating-a-request
add this link to the `sample_trader.cfg` file and the `sample_mirror.cfg` file.

maybe add this link and also headers/params to the `exchanges` command too.

nikhilsaraf:
created this issue from this comment: https://github.com/stellar/kelp/pull/140#issuecomment-482866747

nikhilsaraf:
We would require a code change to support this header parameter since the values in the config file are only used as static headers.

The call stack of the function `plugins.MakeTradingExchange(...)` shows how we go from the botConfig's `EXCHANGE_HEADERS` into static headers on the ccxtExchange (via calling `makeCcxtExchange()` inside `loadExchanges()`).

The `Ccxt` struct (in `ccxt.go`) holds a `headersMap` which is a `map[string]string`.
We can change `headersMap` to be a `map[string]HeaderFn`, where `HeaderFn` is a `func(string, string, string) string` which takes in a `string` representing the http `method`, `requestPath`, and `body` which then produces a header string as an output. This will convert these _static_ headers to be more _dynamic_. 

---

For the existing headers we can just have a passthrough function which ignores the inputs and returns a static header value for the given header key, i.e. when making the TradingExchange and setting the headers, instead of passing an `api.ExchangeHeader` with a fixed string for `Value`, we wrap it in a passthrough function to keep it consistent. Here's a sample of how that can look:
```
api.ExchangeHeader{
    Header: header.Header,
    Value: makeStaticHeaderFn(header.Value),
}

// makeStaticHeaderFn is a convenience method
makeStaticHeaderFn(value string) HeaderFn {
    fn := func(method string, requestPath string, body string) string {
        return value
    }
    // need to convert to HeaderFn to work as a api.ExchangeHeader.Value
    return HeaderFn(fn)
}
```

For the `CB-ACCESS-SIGN` header we can use a function like so:
```
func getCoinbaseSignatureFn(base64EncodedSigningKey string) HeaderFn {
    base64DecodedSigningKey := base64Decode(base64EncodedSigningKey)

    // return this inline method casted as a HeaderFn to work as a headerValue
    return HeaderFn(func(method string, requestPath string, body string) string {
        payload := fmt.Sprintf("%s%s%s%s", time.Now(), method, requestPath, body)
[...]

PR #185, opened 2019-06-16
Add coinbase header reqs to cfg files

gjgollub:
this is done as a resolution of the issue here. Based on the files, it looked to be a simple comment added. Let me know if I misunderstood the requirements. https://github.com/stellar/kelp/issues/156

### DIFF
--- a/gui/web/src/components/molecules/Header/Header.js
+++ b/gui/web/src/components/molecules/Header/Header.js
@@ -15,7 +15,7 @@ class Header extends Component {
               <img src={logo} className={styles.logo} alt="Kelp logo" />
               <span className={styles.version}>{this.props.version}</span>
             </div>
-            <div className={styles.iconMenu}>
+            {/* <div className={styles.iconMenu}>
               <Button
                 icon="day"
                 size="large"
@@ -28,7 +28,7 @@ class Header extends Component {
                 variant="transparent"
                 hsize="round"
               />
-            </div>
+            </div> */}
           </div>
         </div>
       </header>


examples/configs/trader/sample_mirror.cfg
```diff
@@ -64,3 +64,17 @@ PER_LEVEL_SPREAD=0.005
...
 #VALUE=""
+
+# below is a description of coinbase api headers
+# For more detail view the coinebase Api documents: here https://docs.pro.coinbase.com/#creating-a-request
+# ENDPOINT=https://api.pro.coinbase.com
+# The api key as a string.
+#CB-ACCESS-KEY=""
+#The base64-encoded signature (see Signing a Message).
+#timestamp + Method + requestPath + body
+#REQUESTTIME:=time.Now()
```
nikhilsaraf: @gjgollub it looks like `REQUESTTIME` is a dynamic header, i.e. it needs to be set on a per-request basis, is that right?

gjgollub: @nikhilsaraf that's right, it has be done within 30 seconds of the request. Should I note that?

nikhilsaraf: @gjgollub in that case, we would require a code change to support this header parameter since the values in the config file are only used as static headers.

The call stack of the function `plugins.MakeTradingExchange(...)` shows how we go from the botConfig's `EXCHANGE_HEADERS` into static headers on the ccxtExchange (via calling `makeCcxtExchange()` inside `loadExchanges()`).

The `Ccxt` struct (in `ccxt.go`) holds a `headersMap` which is a `map[string]string`.
We can change `headersMap` to be a `map[string]HeaderFn`, where `HeaderFn` is a `func(string, string, string) string` which takes in a `string` representing the http `method`, `requestPath`, and `body` which then produces a header string as an output. This will convert these _static_ headers to be more _dynamic_. 

---

For the existing headers we can just have a passthrough function which ignores the inputs and returns a static header value for the given header key, i.e. when making the TradingExchange and setting the headers, instead of passing an `api.ExchangeHeader` with a fixed string for `Value`, we wrap it in a passthrough function to keep it consistent. Here's a sample of how that can look:
```
api.ExchangeHeader{
    Header: header.Header,
    Value: makeStaticHeaderFn(header.Value),
}

// makeStaticHeaderFn is a convenience method
makeStaticHeaderFn(value string) HeaderFn {
    fn := func(method string, requestPath string, body string) string {
        return value
    }
    // need to convert to HeaderFn to work as a api.ExchangeHeader.Value
    return HeaderFn(fn)
}
```

For the `CB-ACCESS-SIGN` header we can use a function like so:
```
func getCoinbaseSignatureFn(base64EncodedSigningKey string) HeaderFn {
    base64DecodedSigningKey := base64Decode(base64EncodedSigningKey)

    // return this inline method casted as a HeaderFn to work as a headerValue
    return HeaderFn(func(method string, requestPath string, body string) string {
        payload := fmt.Sprintf("%s%s%s%s", time.Now(), method, requestPath, body)
        signature := signMessage(payload, base64DecodedSigningKey)
        base64EncodedHeader := base64Encode(signature)
        return base64EncodedHeader
    })
}
```
This function will be saved as a pre-defined function in a map, pointed to by the string key `COINBASE_FN_CB-ACCCESS-KEY`. We can assign this function on to the api.ExchangeHeader.Value whenever the config file specifies this function.

When setting the headers via the config we can specify the function to use followed by the required params, so for example:
```
CB-ACCCESS-KEY="STATIC:thisismykey"
CB-ACCCESS-KEY="COINBASE_FN_CB-ACCCESS-KEY:mycoinbasesigningkey"
```

This is a fairly more involved change than you may have anticipated given that it requires a code change and small refactor because of the dynamic nature of one of the Coinbase headers. Let me know if you are still up to making this change and if you have any questions about this. Happy to get on a call if that will be helpful.

gjgollub: Sure, I'll give it a try

nikhilsaraf: great, thanks!

nikhilsaraf:
@gjgollub any updates on this PR?

gjgollub:
@nikhilsaraf yes, I started work on it, but I won't have time to look at it again until this weekend. I haven't forgotten about it! Thanks for the patience

nikhilsaraf:
@gjgollub Ok, not a problem, and thank you for the update.
checking in to see if I could help with anything so please let me know if something comes up or if you want an additional details on the issue.

gjgollub:
Hi @nikhilsaraf, do you think we can set up some time to hop on a call?

nikhilsaraf:
Hi @gjgollub yes, could you ping me on keybase so we can figure out the details of the call? My keybase handle is https://keybase.io/nikhilsaraf
(we use keybase chat for discussions in the stellar community)

gjgollub:
Message sent on Keybase

PR #205, opened 2019-07-25
Use txnbuild

poliha:
This PR covers part of the work required to update Kelp to use the new stellar Go sdk. This PR focuses on changing the transaction builder package from `build` to `txnbuild`.

This PR also changes the usage of `clients/horizon` in some places that could not be changed in #204. In particular, these two files were changed

.../gui/backend/upsert_bot_config.go
.../gui/backend/autogenerate_bot.go

With these changes, Kelp should not have any instance of the deprecated `build` and `clients/horizon` packages.
This PR is based off the branch in #204, please review that before this. 

------
Nikhil's list of outstanding issues to check:
- [x] orderConstraintsFilter logs `baseAmount` and `quoteAmount` numbers that are divided by 10^7. Example: `0.00015000` instead of `1500.0000000` when the other log lines use the correct value of `1500.0000000`
- [x] revert `txnbuild.Operation` references in `api/*` so we don't bump major version number and create a new PR for this API change to be merged later


plugins/makerModeFilter.go
```diff
@@ -132,21 +133,16 @@ func (f *makerModeFilter) filterOps(
...
+		var newOp txnbuild.Operation
 		var keep bool
 		switch o := op.(type) {
-		case *build.ManageOfferBuilder:
+		case *txnbuild.ManageSellOffer:
 			newOp, keep, e = f.transformOfferMakerMode(baseAsset, quoteAsset, topBidPrice, topAskPrice, o)
 			if e != nil {
 				return nil, fmt.Errorf("could not transform offer (pointer case): %s", e)
 			}
-		case build.ManageOfferBuilder:
```
nikhilsaraf: I think we need this too - sometimes the value was not a pointer. is that not possible with `txnbuild.Operation`?

plugins/mirrorStrategy.go
```diff
@@ -346,7 +346,7 @@ func (s *mirrorStrategy) updateLevels(
 				return nil, e
 			}
 			if mo != nil {
-				ops = append(ops, *mo)
+				ops = append(ops, mo)
```
nikhilsaraf: here too -- can `mo` ever be a pointer?

plugins/mirrorStrategy.go
```diff
@@ -375,7 +375,7 @@ func (s *mirrorStrategy) updateLevels(
 		// delete remaining prior offers
 		for i := len(newOrders); i < len(oldOffers); i++ {
 			deleteOp := s.sdex.DeleteOffer(oldOffers[i])
-			deleteOps = append(deleteOps, deleteOp)
+			deleteOps = append(deleteOps, &deleteOp)
```
nikhilsaraf: we are appending a reference to the delete-op here, why not use the deleteOp directly?
(when comparing to the usage of `mo` is it because the `mo` variable is already a pointer and we're only ever using pointers for the `txnbuild.Operation`?)

plugins/orderConstraintsFilter.go
```diff
@@ -34,51 +35,50 @@ func MakeFilterOrderConstraints(
...
+			if opPtr.OfferID == 0 {
 				// new offers can be dropped, so don't add to filteredOps
-			} else if opPtr.MO.Amount != 0 {
+			} else if amountFloat != 0 {
 				// modify offers should be converted to delete offers
 				opCopy := *opPtr
-				opCopy.MO.Amount = 0
-				filteredOps = append(filteredOps, opCopy)
+				opCopy.Amount = "0"
+				filteredOps = append(filteredOps, &opCopy)
```
nikhilsaraf: why do we use a pointer of `opCopy` here whereas on the left hand side of this diff we are passing the underlying struct?

plugins/batchedExchange.go
```diff
@@ -363,27 +363,34 @@ func convert2Price(number *model.Number) (hProtocol.Price, error) {
...
-	price := model.NumberFromFloat(float64(mob.MO.Price.N)/float64(mob.MO.Price.D), largePrecision)
-	volume := model.NumberFromFloat(float64(mob.MO.Amount)/math.Pow(10, 7), largePrecision)
-	isBuy, e := assetsEqual(quoteAsset, mob.MO.Selling)
+
+	priceFloat, e := strconv.ParseFloat(mob.Price, 64)
+	if e != nil {
+		return nil, fmt.Errorf("could not convert price to float: %s", e)
+	}
+
+	amountFloat, e := strconv.ParseFloat(mob.Amount, 64)
```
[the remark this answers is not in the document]
poliha: Amount here is float, but represented as a string. Does this have to be stroops?

plugins/batchedExchange.go
```diff
@@ -363,27 +363,34 @@ func convert2Price(number *model.Number) (hProtocol.Price, error) {
...
+	if e != nil {
+		return nil, fmt.Errorf("could not convert amount to float: %s", e)
+	}
+
+	price := model.NumberFromFloat(priceFloat, largePrecision)
+	volume := model.NumberFromFloat(amountFloat/math.Pow(10, 7), largePrecision)
```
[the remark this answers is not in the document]
poliha: hmm we probably dont need the division here if it is float

plugins/makerModeFilter.go
poliha: This will always be a pointer because the structs that implement the `txnbuild.Operation` interface are pointers

plugins/mirrorStrategy.go
poliha: here `mo` is already a pointer.

plugins/mirrorStrategy.go
poliha: Yes that is correct. this is similar to `mo` above

plugins/orderConstraintsFilter.go
poliha: `filteredOps` is an array of `txnbuild.Operation` and it can only be pointers. Same behaviour as `mo` above

poliha:
No worries, please take your time. It is essential that the changes do not introduce any bugs so it is a good call to delay on this a bit till you have more bandwidth.

plugins/orderConstraintsFilter.go
```diff
@@ -34,51 +35,50 @@ func MakeFilterOrderConstraints(
...
-			keep, e = f.shouldKeepOffer(&o)
-			if e != nil {
-				return nil, fmt.Errorf("could not check transform offer (non-pointer case): %s", e)
-			}
-			opPtr = &o
 		default:
 			keep = true
 		}
 
+		amountFloat, e := strconv.ParseFloat(opPtr.Amount, 64)
```
nikhilsaraf: note to self, check amount is in stroops or not

plugins/orderConstraintsFilter.go
```diff
@@ -87,20 +87,28 @@ func (f *orderConstraintsFilter) Apply(
 	return filteredOps, nil
 }
 
-func (f *orderConstraintsFilter) shouldKeepOffer(op *build.ManageOfferBuilder) (bool, error) {
+func (f *orderConstraintsFilter) shouldKeepOffer(op *txnbuild.ManageSellOffer) (bool, error) {
 	// delete operations should never be dropped
-	if op.MO.Amount == 0 {
+	amountFloat, e := strconv.ParseFloat(op.Amount, 64)
```
nikhilsaraf: note to self, amount conversion is in stroops?

plugins/sdex.go
```diff
@@ -338,60 +340,59 @@ func (sdex *SDEX) createModifySellOffer(offer *hProtocol.Offer, selling hProtoco
 	}
 
 	stringPrice := strconv.FormatFloat(price, 'f', int(sdexOrderConstraints.PricePrecision), 64)
-	rate := build.Rate{
-		Selling: utils.Asset2Asset(selling),
-		Buying:  utils.Asset2Asset(buying),
-		Price:   build.Price(stringPrice),
-	}
+	stringAmount := strconv.FormatFloat(amount, 'f', int(sdexOrderConstraints.VolumePrecision), 64)
```
nikhilsaraf: note to self, check that stringAmount should not be in stroops here

plugins/makerModeFilter.go
```diff
@@ -178,19 +174,29 @@ func (f *makerModeFilter) transformOfferMakerMode(
 	quoteAsset hProtocol.Asset,
 	topBidPrice *model.Number,
 	topAskPrice *model.Number,
-	op *build.ManageOfferBuilder,
-) (*build.ManageOfferBuilder, bool, error) {
+	op *txnbuild.ManageSellOffer,
+) (*txnbuild.ManageSellOffer, bool, error) {
+	//represent amount as float
+	amountFloat, e := strconv.ParseFloat(op.Amount, 64)
```
nikhilsaraf: note to self: check float conversion

plugins/makerModeFilter.go
```diff
@@ -215,14 +221,14 @@ func (f *makerModeFilter) transformOfferMakerMode(
...
-	} else if op.MO.Amount != 0 {
+	} else if amountFloat != 0 {
 		// modify offers should be converted to delete offers
 		opCopy := *op
-		opCopy.MO.Amount = 0
+		opCopy.Amount = "0"
 		return &opCopy, false, nil
 	}
-	return nil, keep, fmt.Errorf("unable to transform manageOffer operation: offerID=%d, amount=%.7f, price=%.7f", op.MO.OfferId, float64(op.MO.Amount)/math.Pow(10, 7), sellPrice)
+	return nil, keep, fmt.Errorf("unable to transform manageOffer operation: offerID=%d, amount=%.7f, price=%.7f", op.OfferID, amountFloat/math.Pow(10, 7), sellPrice)
```
nikhilsaraf: note to self, check float display

plugins/batchedExchange.go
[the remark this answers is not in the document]
nikhilsaraf: no, it should be as a float value

plugins/batchedExchange.go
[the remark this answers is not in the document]
nikhilsaraf: fixed.

plugins/makerModeFilter.go
nikhilsaraf: fixed

plugins/makerModeFilter.go
nikhilsaraf: not needed, removed

plugins/orderConstraintsFilter.go
nikhilsaraf: don't need amountFloat here since we only check for "0" value.. fixed.

plugins/orderConstraintsFilter.go
nikhilsaraf: no, as a float value

plugins/orderConstraintsFilter.go
```diff
@@ -87,20 +87,28 @@ func (f *orderConstraintsFilter) Apply(
...
 
-	sellPrice := float64(op.MO.Price.N) / float64(op.MO.Price.D)
+	sellPrice, e := strconv.ParseFloat(op.Price, 64)
+	if e != nil {
+		return false, fmt.Errorf("could not convert price to float: %s", e)
+	}
+
 	if isSell {
-		baseAmount := float64(op.MO.Amount) / math.Pow(10, 7)
+		baseAmount := amountFloat / math.Pow(10, 7)
```
nikhilsaraf: doesn't need the division here... removed

plugins/orderConstraintsFilter.go
```diff
@@ -115,7 +123,7 @@ func (f *orderConstraintsFilter) shouldKeepOffer(op *build.ManageOfferBuilder) (
 	}
 
 	// buying
-	quoteAmount := float64(op.MO.Amount) / math.Pow(10, 7)
+	quoteAmount := amountFloat / math.Pow(10, 7)
```
nikhilsaraf: doesn't need division here either... removed

Issue #248, opened 2019-08-16
Build Compatibility for Git Bash For Windows


jcperkins12:
## Describe the bug
Continuation of [Issue#14](https://github.com/stellar/kelp/issues/14) which seemed to be closed by insisting on building with a BASH console. 

The bug is that I still get the "need to invoke from the root 'kelp' directory" error when trying to build the project on the Git Bash console on my windows computer. 

## Expected behavior
I think the correct behavior is to be able to build the program with the bash console that is packaged with Git

## Frequency
The frequency is always

## Steps To Reproduce
Here are the steps to reproduce the issue (see attachments in section below):
1. On a windows 10 machine download git and include the git bash console
2. Follow the build from source instructions on the readme with the bash console
3. When you I try `./scripts/build.sh` the script exits with the message "need to invoke from the root 'kelp' directory" even though I have already moved into the 'kelp' directory per the instructions. 
![image](https://user-images.githubusercontent.com/37156131/63181896-d2cd6380-c016-11e9-9f52-ba93f9af8ba5.png)

## Possible Solution
After some research ([Git bash doesn't recognize 'rev' command](https://stackoverflow.com/q/55937274/5477004)), it appears that "rev is an external command (a separate executable), not part of the shell itself; it's thus not part of any version of bash, but something that exists (or doesn't) depending on whether your it's been separately installed (or included with your operating system &c)". 

As mentioned [Issue#14 comment](https://github.com/stellar/kelp/issues/14#issuecomment-426487056) the problem is in the line `pwd | rev | cut -d'/' -f1 | rev` which can be rewritten as either ([more research](https://stackoverflow.com/a/1744613/5477004))
```$(basename $("pwd"))```
or 
```pwd | sed 's#.*/##'```

Both have worked on my system and I have little linux experience to known whether either will cause issues in other environments. With that being said, my preference is for the first solution as I find it more readable.

## Your Environment
When successfully run my environment is:
![image](https://user-images.githubusercontent.com/37156131/63181986-04462f00-c017-11e9-8c8f-64c17f21561a.png)

## Context
As a predominantly windows user, I have found the git bash console to be the easiest one to use out of the box and is the most user friendly for those unfamiliar with unix. In order to clone the kelp repo, most users will already have git installed and likely already have this console. As such and baring integration issues with other systems, it makes sense to to try to restrict the bash scripts to commands that are native to bash whenever possible.

PR #249, opened 2019-08-16
git bash windows compatibility

jcperkins12:
This PR address [Issue#248](https://github.com/stellar/kelp/issues/248#issue-481691526).

It replaces the `rev` command which may be common, but not native to bash, with a native bash command that works on the out of the box bash console for windows that can be installed alongside Git.


jcperkins12:
@nikhilsaraf  thank you for looking into it. I thought that I had signed the agreement before I created the pull request. Are you able to tell if it is on record now? If so, is there anything else that you need from me and if not, I can fill out another one, but is there something I need to do differently? The original agreement was associated with the email address jcperkins12@gmail.com

PR #380, opened 2020-02-20
Build new offer using private helper instead of DeleteOfferOp2.

abuiles:
Instead of adding `DeleteOfferOp2` which seems to be required only by Kelp at the moment, I'd like us to us a private helper method inside of Kelp to produce the same result.

This will unblock Kelp and prevent us from ending up in the SDK which a function which is custom made for Kelp.

Nikhil, can you open an issue in the go sdk and explain again what your initial use case was and then maybe propose a change on `DeleteOfferOp` -- that way we can add it on the next major release.


nikhilsaraf:
will make this change once I migrate over to v2.1 of the horizonclient, thanks for the PR and will use it as reference

PR #404, opened 2020-04-16
Kelp UI: Fix basepath and use pingURL to ping server in tailFile before redirect

nikhilsaraf:


nikhilsaraf:
circleci is failing because of rate limiting on the circleci instance (ran on master and it failed too after master was already passing)

PR #492, opened 2020-08-28
Increase user-friendliness of short of assets error (closes #104)

### LINKED ISSUE #104: If short of assets, the error message should be a lot more user friendly
// see a sample bug report here: https://github.com/interstellar/kelp/issues/1

## Describe the bug
// A succinct description of the buggy behavior.
// This section includes **what** the bot is doing and **why** you think it is incorrect.

kelp log says:
Feb 07 06:12:46 steve-kelp bash[2072]: 2019/02/07 06:12:46 error in selling sub-strategy: unable to create preceding offers: unable to create new preceding offer: error: (programmer?) unable to create offer but available capacities were more than the attempted offer amounts, sellingCapacity=77297.7136387, incrementalSellAmount=5.0000000, buying
Capacity=17976931348623157081452742373170435679807056752584499659891747680315726078002853876058955863276687817154045895351438246423432132688946418276846754670353751698604991
0576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368.0000000, incrementalBuyAmount=68.6164885

## Expected behavior
// A succinct description of **what you expected to happen**.

Just tell me I need to put more of asset XXX in the account to be able to cover the orders!

## Frequency
// **How often** does this happen?
//     Unpredictable = it's hard to predict when it will happen
//     Sometimes = it happens every now and then but there's no pattern to it
//     Always = it always happens consistently at the same point

The frequency is: ...

## Steps To Reproduce
// A concise description of the **steps to reproduce the buggy behavior** along with screenshots wherever applicable:

Here are the steps to reproduce the issue (see attachments in section below):
1. Set up your configuration file with these parameters: ...
2. Start the bot using this command: ...
...

## Possible Solution
// Optional, suggest a fix.

This bug can be fixed by ...

## Your Environment
// Run `kelp version` to get the version and build information and paste it here
//   version: v1.0.0-rc1
//   git hash: 1bcc4e9fa4012867a549472dbdf61cb008041305
//   build date: 20180813T013016Z
//   GOOS: linux
//   GOARCH: amd64

...

## Context
// A succinct description of how has this bug has **affected you** or prevented you from accomplishing what you wanted.

I am unable to ... because of this bug.

## Attachments
// Attach any relevant configuration files, logs, tx hashes, etc. here.

debnil:
This PR increases the user-friendliness of the error when short of assets. This closes #104.

### DIFF
--- a/plugins/sellSideStrategy.go
+++ b/plugins/sellSideStrategy.go
@@ -351,7 +351,7 @@ func (s *sellSideStrategy) computeRemainderAmount(incrementalSellAmount float64,
 	}
 
 	if availableSellingCapacity.Selling >= incrementalSellAmount && availableBuyingCapacity.Buying >= incrementalBuyAmount {
-		return 0, 0, fmt.Errorf("error: (programmer?) unable to create offer but available capacities were more than the attempted offer amounts, sellingCapacity=%.8f, incrementalSellAmount=%.8f, buyingCapacity=%.8f, incrementalBuyAmount=%.8f",
+		return 0, 0, fmt.Errorf("error: could not create offer, but available capacity exceeds the offer amounts; to fix, increase the offer target amount; sellingCapacity=%.8f, incrementalSellAmount=%.8f, buyingCapacity=%.8f, incrementalBuyAmount=%.8f",
 			availableSellingCapacity.Selling, incrementalSellAmount, availableBuyingCapacity.Buying, incrementalBuyAmount)
 	}
 


plugins/sellSideStrategy.go
```diff
@@ -351,7 +351,7 @@ func (s *sellSideStrategy) computeRemainderAmount(incrementalSellAmount float64,
 	}
 
 	if availableSellingCapacity.Selling >= incrementalSellAmount && availableBuyingCapacity.Buying >= incrementalBuyAmount {
-		return 0, 0, fmt.Errorf("error: (programmer?) unable to create offer but available capacities were more than the attempted offer amounts, sellingCapacity=%.8f, incrementalSellAmount=%.8f, buyingCapacity=%.8f, incrementalBuyAmount=%.8f",
```
nikhilsaraf, line 354: I think this old error message is correct. See the if condition -- it should always hold true. If it is false then it means something has gone terribly wrong and it's likely a programmer error as indicated by the error message.

IIRC this error has always occurred whenever there were less assets in the user's account so this error message was only a red herring.

There is another error message that indicates that the user has less assets. Search for the following in `ieif.go`:
`we will oversell`

There are two occurrences of that message, we should leave those two error messages as-is and add this additional line to hint to the user the corrective action they need to take:
```
utils.PrintErrorHintf("Your account may have run out of the asset '%s'. Add more tokens of '%s' so the bot can run correctly")
```

You will need to reproduce this issue to confirm that it is working correctly. Once fixed, could you please upload the section of the log file (between the two separator lines `--------`) when the user has a low balance for a bot trading (a) SOME_ASSET_A vs XLM where XLM is low; and (b) market trading with SOME_ASSET_A vs. SOME_ASSET_B where XLM is low and it is unable to pay fees?

nikhilsaraf (changes requested):
let me know if you have any questions.

debnil:
@nikhilsaraf: Got it, thanks!! Would you be able to write e steps to reproduce the ieif.go error on the original issue, just to make sure we're on the same page?

nikhilsaraf:
Added some details on the issue. Let me know if that helps or we can discuss on Monday too.

debnil:
Those are super helpful. I'm just going to close this PR - the changes are super different, and this way, I can contribute from a fork rather than a branch pushed to `master`.

PR #513, opened 2020-09-25
Set up CLI data tracking

### LINKED ISSUE #478: [8] Improve data tracking for Kelp CLI
This is dependent on the outcome from the Spike https://github.com/stellar/kelp/issues/477

Outcome is to use Amplitude

See this doc for details on what to track: https://docs.google.com/document/d/1R5hU7dSloQHZ0XNZkUrS0LPOr2gLVpteXq5NvR-Fy2k

Follow-on issues to be handled as separate tasks: https://github.com/stellar/kelp/issues/516

debnil:
**What/Why**
This PR sets up Amplitude events for setup, update, and delete events. This will help enable data tracking for Kelp. This closes #478.

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -505,11 +505,11 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	// Fail if in release mode with undefined API key.
+	// TODO: Fail if in release mode with undefined API key.
 
 	userID := "12345" // TODO: Properly generate and save user ID.
 	httpClient := &http.Client{}
-	tracker, e := metrics.MakeMetricsTracker(
+	metricsTracker, e := metrics.MakeMetricsTracker(
 		userID,
 		amplitudeAPIKey,
 		httpClient,
@@ -528,7 +528,7 @@ func runTradeCmd(options inputs) {
 		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker with error: %s", e))
 	}
 
-	e = tracker.SendStartupEvent()
+	e = metricsTracker.SendStartupEvent()
 	if e != nil {
 		logger.Fatal(l, fmt.Errorf("could not send startup event metric with error: %s", e))
 	}
@@ -649,7 +649,7 @@ func runTradeCmd(options inputs) {
 		options,
 		threadTracker,
 		db,
-		tracker,
+		metricsTracker,
 	)
 	fillTracker := makeFillTracker(
 		l,
@@ -663,7 +663,7 @@ func runTradeCmd(options inputs) {
 		db,
 		threadTracker,
 		botConfig.DbOverrideAccountID,
-		tracker,
+		metricsTracker,
 	)
 	bot := makeBot(
 		l,
@@ -678,7 +678,7 @@ func runTradeCmd(options inputs) {
 		fillTracker,
 		threadTracker,
 		options,
-		tracker,
+		metricsTracker,
 	)
 	// --- end initialization of objects ---
 	// --- start initialization of services ---
@@ -693,7 +693,7 @@ func runTradeCmd(options inputs) {
 				// we want to delete all the offers and exit here because we don't want the bot to run if monitoring isn't working
 				// if monitoring is desired but not working properly, we want the bot to be shut down and guarantee that there
 				// aren't outstanding offers.
-				deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker, tracker)
+				deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker, metricsTracker)
 			}
 		}()
 	}
@@ -705,7 +705,7 @@ func runTradeCmd(options inputs) {
 				l.Info("")
 				l.Errorf("problem encountered while running the fill tracker: %s", e)
 				// we want to delete all the offers and exit here because we don't want the bot to run if fill tracking isn't working
-				deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker, tracker)
+				deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker, metricsTracker)
 			}
 		}()
 	}
@@ -862,7 +862,7 @@ func deleteAllOffersAndExit(
 	sdex *plugins.SDEX,
 	exchangeShim api.ExchangeShim,
 	threadTracker *multithreading.ThreadTracker,
-	tracker *metrics.MetricsTracker,
+	metricsTracker *metrics.MetricsTracker,
 ) {
 	l.Info("")
 	l.Infof("waiting for all outstanding threads (%d) to finish before loading offers to be deleted...", threadTracker.NumActiveThreads())
@@ -898,11 +898,12 @@ func deleteAllOffersAndExit(
 			return
 		}
 
-		e = tracker.SendDeleteEvent(true)
-		if e != nil {
-			logger.Fatal(l, fmt.Errorf("could not send delete event metric with error: %s", e))
-			return
-		}
+		threadTracker.TriggerGoroutine(func(inputs []interface{}) {
+			e := metricsTracker.SendDeleteEvent(true)
+			if e != nil {
+				logger.Fatal(l, fmt.Errorf("could not send delete event metric with error: %s", e))
+			}
+		}, nil)
 
 		for {
 			sleepSeconds := 10
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -505,6 +505,8 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
+	// Fail if in release mode with undefined API key.
+
 	userID := "12345" // TODO: Properly generate and save user ID.
 	httpClient := &http.Client{}
 	tracker, e := metrics.MakeMetricsTracker(
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -333,7 +333,7 @@ func makeStrategy(
 	options inputs,
 	threadTracker *multithreading.ThreadTracker,
 	db *sql.DB,
-	tracker *metrics.Tracker,
+	tracker *metrics.MetricsTracker,
 ) api.Strategy {
 	// setting the temp hack variables for the sdex price feeds
 	e := plugins.SetPrivateSdexHack(client, plugins.MakeIEIF(true), network)
@@ -382,7 +382,7 @@ func makeBot(
 	fillTracker api.FillTracker,
 	threadTracker *multithreading.ThreadTracker,
 	options inputs,
-	metricsTracker *metrics.Tracker,
+	metricsTracker *metrics.MetricsTracker,
 ) *trader.Trader {
 	timeController := plugins.MakeIntervalTimeController(
 		time.Duration(botConfig.TickIntervalSeconds)*time.Second,
@@ -505,19 +505,30 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	tracker := metrics.MakeMetricsTracker(
+	userID := "12345" // TODO: Properly generate and save user ID.
+	httpClient := &http.Client{}
+	tracker, e := metrics.MakeMetricsTracker(
+		userID,
+		amplitudeAPIKey,
+		httpClient,
+		start,
 		version,
 		runtime.GOOS,
-		*options.ui,
+		runtime.GOARCH,
+		"", // TODO: Determine how to get GOARM.
+		guiVersion,
 		*options.strategy,
-		float64(botConfig.TickIntervalSeconds),
+		botConfig.TickIntervalSeconds,
 		botConfig.TradingExchange,
-		fmt.Sprintf("%s:%s/%s:%s", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB),
+		botConfig.TradingPair(),
 	)
-
-	e := tracker.SendStartupEvent(l)
 	if e != nil {
-		l.Info("could not send amplitude event")
+		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker with error: %s", e))
+	}
+
+	e = tracker.SendStartupEvent()
+	if e != nil {
+		logger.Fatal(l, fmt.Errorf("could not send startup event metric with error: %s", e))
 	}
 
 	// --- start initialization of objects ----
@@ -754,7 +765,7 @@ func makeFillTracker(
 	db *sql.DB,
 	threadTracker *multithreading.ThreadTracker,
 	accountID string,
-	metricsTracker *metrics.Tracker,
+	metricsTracker *metrics.MetricsTracker,
 ) api.FillTracker {
 	strategyFillHandlers, e := strategy.GetFillHandlers()
 	if e != nil {
@@ -849,7 +860,7 @@ func deleteAllOffersAndExit(
 	sdex *plugins.SDEX,
 	exchangeShim api.ExchangeShim,
 	threadTracker *multithreading.ThreadTracker,
-	tracker *metrics.Tracker,
+	tracker *metrics.MetricsTracker,
 ) {
 	l.Info("")
 	l.Infof("waiting for all outstanding threads (%d) to finish before loading offers to be deleted...", threadTracker.NumActiveThreads())
@@ -885,9 +896,9 @@ func deleteAllOffersAndExit(
 			return
 		}
 
-		e = tracker.SendDeleteEvent(l, true)
+		e = tracker.SendDeleteEvent(true)
 		if e != nil {
-			l.Info(fmt.Sprintf("Sending deletion data event failed with error: %v\n", e))
+			logger.Fatal(l, fmt.Errorf("could not send delete event metric with error: %s", e))
 			return
 		}
 
@@ -915,12 +926,15 @@ func setLogFile(l logger.Logger, filename string) {
 	defer logPanic(l, false)
 }
 
+var start time.Time
+
 func makeLogFilename(logPrefix string, botConfig trader.BotConfig) string {
-	t := time.Now().Format("20060102T150405MST")
+	start = time.Now()
+	startStr := start.Format("20060102T150405MST")
 	if botConfig.IsTradingSdex() {
-		return fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, t)
+		return fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, startStr)
 	}
-	return fmt.Sprintf("%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.AssetCodeB, t)
+	return fmt.Sprintf("%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.AssetCodeB, startStr)
 }
 
 func parseValueFeed(valueFeed string) (api.PriceFeed, error) {
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -199,7 +199,7 @@ func makeFeeFn(l logger.Logger, botConfig trader.BotConfig, newClient *horizoncl
 	return feeFn
 }
 
-func readBotConfig(l logger.Logger, options inputs) trader.BotConfig {
+func readBotConfig(l logger.Logger, options inputs, botStart time.Time) trader.BotConfig {
 	var botConfig trader.BotConfig
 	e := config.Read(*options.botConfigPath, &botConfig)
 	utils.CheckConfigError(botConfig, e, *options.botConfigPath)
@@ -209,7 +209,7 @@ func readBotConfig(l logger.Logger, options inputs) trader.BotConfig {
 	}
 
 	if *options.logPrefix != "" {
-		logFilename := makeLogFilename(*options.logPrefix, botConfig)
+		logFilename := makeLogFilename(*options.logPrefix, botConfig, botStart)
 		setLogFile(l, logFilename)
 	}
 
@@ -501,36 +501,42 @@ func convertDeprecatedBotConfigValues(l logger.Logger, botConfig trader.BotConfi
 
 func runTradeCmd(options inputs) {
 	l := logger.MakeBasicLogger()
-	botConfig := readBotConfig(l, options)
+	botStart := time.Now()
+	botConfig := readBotConfig(l, options, botStart)
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	// TODO: Fail if in release mode with undefined API key.
+	// TODO DS Fail if in release mode with undefined API key.
 
-	userID := "12345" // TODO: Properly generate and save user ID.
+	userID := "-1" // TODO DS Properly generate and save user ID.
 	httpClient := &http.Client{}
+	var guiVersionFlag string
+	if *options.ui {
+		guiVersionFlag = guiVersion
+	}
+
 	metricsTracker, e := metrics.MakeMetricsTracker(
 		userID,
 		amplitudeAPIKey,
 		httpClient,
-		start,
+		botStart,
 		version,
 		runtime.GOOS,
 		runtime.GOARCH,
-		"", // TODO: Determine how to get GOARM.
-		guiVersion,
+		"", // TODO DS Determine how to get GOARM.
+		guiVersionFlag,
 		*options.strategy,
 		botConfig.TickIntervalSeconds,
 		botConfig.TradingExchange,
 		botConfig.TradingPair(),
 	)
 	if e != nil {
-		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker with error: %s", e))
+		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker: %s", e))
 	}
 
 	e = metricsTracker.SendStartupEvent()
 	if e != nil {
-		logger.Fatal(l, fmt.Errorf("could not send startup event metric with error: %s", e))
+		logger.Fatal(l, fmt.Errorf("could not send startup event metric: %s", e))
 	}
 
 	// --- start initialization of objects ----
@@ -864,6 +870,14 @@ func deleteAllOffersAndExit(
 	threadTracker *multithreading.ThreadTracker,
 	metricsTracker *metrics.MetricsTracker,
 ) {
+	// synchronous event to guarantee execution. we want to know whenever we enter the delete all offers logic. this function
+	// waits for all threads to be synchronous, which is equivalent to sending synchronously. we use
+	e := metricsTracker.SendDeleteEvent(true)
+	if e != nil {
+		// We don't want to crash upon failure, so offers will be deleted regardless of metric send.
+		l.Infof("could not send delete event metric: %s", e)
+	}
+
 	l.Info("")
 	l.Infof("waiting for all outstanding threads (%d) to finish before loading offers to be deleted...", threadTracker.NumActiveThreads())
 	threadTracker.Stop(multithreading.StopModeError)
@@ -898,13 +912,6 @@ func deleteAllOffersAndExit(
 			return
 		}
 
-		threadTracker.TriggerGoroutine(func(inputs []interface{}) {
-			e := metricsTracker.SendDeleteEvent(true)
-			if e != nil {
-				logger.Fatal(l, fmt.Errorf("could not send delete event metric with error: %s", e))
-			}
-		}, nil)
-
 		for {
 			sleepSeconds := 10
 			l.Infof("sleeping for %d seconds until our deletion is confirmed and we exit...(should never reach this line since we submit delete ops synchronously)\n", sleepSeconds)
@@ -929,15 +936,12 @@ func setLogFile(l logger.Logger, filename string) {
 	defer logPanic(l, false)
 }
 
-var start time.Time
-
-func makeLogFilename(logPrefix string, botConfig trader.BotConfig) string {
-	start = time.Now()
-	startStr := start.Format("20060102T150405MST")
+func makeLogFilename(logPrefix string, botConfig trader.BotConfig, botStart time.Time) string {
+	botStartStr := botStart.Format("20060102T150405MST")
 	if botConfig.IsTradingSdex() {
-		return fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, startStr)
+		return fmt.Sprintf("%s_%s_%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB, botStartStr)
 	}
-	return fmt.Sprintf("%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.AssetCodeB, startStr)
+	return fmt.Sprintf("%s_%s_%s_%s.log", logPrefix, botConfig.AssetCodeA, botConfig.AssetCodeB, botStartStr)
 }
 
 func parseValueFeed(valueFeed string) (api.PriceFeed, error) {
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"runtime"
 	"runtime/debug"
 	"strings"
 	"time"
@@ -23,6 +24,7 @@ import (
 	"github.com/stellar/kelp/plugins"
 	"github.com/stellar/kelp/support/database"
 	"github.com/stellar/kelp/support/logger"
+	"github.com/stellar/kelp/support/metrics"
 	"github.com/stellar/kelp/support/monitoring"
 	"github.com/stellar/kelp/support/networking"
 	"github.com/stellar/kelp/support/prefs"
@@ -331,6 +333,7 @@ func makeStrategy(
 	options inputs,
 	threadTracker *multithreading.ThreadTracker,
 	db *sql.DB,
+	tracker *metrics.Tracker,
 ) api.Strategy {
 	// setting the temp hack variables for the sdex price feeds
 	e := plugins.SetPrivateSdexHack(client, plugins.MakeIEIF(true), network)
@@ -338,7 +341,7 @@ func makeStrategy(
 		l.Info("")
 		l.Errorf("%s", e)
 		// we want to delete all the offers and exit here since there is something wrong with our setup
-		deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker)
+		deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker, tracker)
 	}
 
 	strategy, e := plugins.MakeStrategy(
@@ -361,7 +364,7 @@ func makeStrategy(
 		l.Info("")
 		l.Errorf("%s", e)
 		// we want to delete all the offers and exit here since there is something wrong with our setup
-		deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker)
+		deleteAllOffersAndExit(l, botConfig, client, sdex, exchangeShim, threadTracker, tracker)
 	}
 	return strategy
 }
@@ -379,6 +382,7 @@ func makeBot(
 	fillTracker api.FillTracker,
 	threadTracker *multithreading.ThreadTracker,
 	options inputs,
+	metricsTracker *metrics.Tracker,
 ) *trader.Trader {
 	timeController := plugins.MakeIntervalTimeController(
 		time.Duration(botConfig.TickIntervalSeconds)*time.Second,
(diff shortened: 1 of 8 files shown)


debnil:
**Known Limitations**
There are several known limitations that I need to figure out, hence this is a draft PR.

1. How do I read in the Amplitude API key from the environment? Currently, this doesn't work unless I hardcode the Amplitude API key in `support/metricsTracker.go` when I define `amplitudeAPIKey`. I can't read it from the environment defined in `scripts/build.sh`. Additionally, if I define the `AMPLITUDE_API_KEY` env var locally, I cannot read this variable through `echo AMPLITUDE_API_KEY` in `scripts/build.sh`. @nikhilsaraf I feel like you have a vision for how to do this in Kelp, so let me know.
2. How do I test the deletion event? I can successfully test the startup and update events by running `kelp trade` with any example strategy. What command should I run to confirm that the deletion event works?
3. Is there a `deleteOffers` function? I'm calling the Amplitude event within `deleteOffersAndExit`, but I can't find a comparable function for `deleteOffers`.

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+
+	defer resp.Body.Close()
+
+	body, e := ioutil.ReadAll(resp.Body)
+	if e != nil {
+		l.Info("could not read response body")
+		return e
+	}
+
+	l.Info(fmt.Sprintf("Successfully sent startup event with response %v\n", string(body)))
```
nikhilsaraf, line 163: we don't want to log the response since we don't really care about it. What we care about is the event we sent. This not always the startup event, so we should log the `eventName`

cmd/trade.go
```diff
@@ -331,14 +333,15 @@ func makeStrategy(
 	options inputs,
 	threadTracker *multithreading.ThreadTracker,
 	db *sql.DB,
+	tracker *metrics.Tracker,
```
nikhilsaraf, line 336: let's call this `metricsTracker` since this is our third type of tracker in the codebase

cmd/trade.go
```diff
@@ -500,6 +505,21 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
+	tracker := metrics.MakeMetricsTracker(
+		version,
+		runtime.GOOS,
+		*options.ui,
```
nikhilsaraf, line 511: since we've changed the logic for `ui` (see comment in metricsTracker) we should pass the `guiVersion` if `*options.ui` is enabled

cmd/trade.go
```diff
@@ -500,6 +505,21 @@ func runTradeCmd(options inputs) {
...
+		*options.strategy,
+		float64(botConfig.TickIntervalSeconds),
```
nikhilsaraf, line 513: should be an integer

cmd/trade.go
```diff
@@ -500,6 +505,21 @@ func runTradeCmd(options inputs) {
...
+		botConfig.TradingExchange,
+		fmt.Sprintf("%s:%s/%s:%s", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB),
```
nikhilsaraf, line 515: this tradingPair is not correct. if `botConfig.IsTradingSdex()` then this is correct, else it's just `codeA/codeB` (i.e. no issuers).

not sure if this logic is specified elsewhere in the codebase (probably is) but if you can't find it then it's ok to put this in a separate function on `botConfig` struct so we can reuse it later

cmd/trade.go
```diff
@@ -500,6 +505,21 @@ func runTradeCmd(options inputs) {
...
+	)
+
+	e := tracker.SendStartupEvent(l)
+	if e != nil {
+		l.Info("could not send amplitude event")
```
nikhilsaraf, line 520: it's generally a bad practice to eat up errors. if there is a need to eat up errors then it should have a comment next to it explaining why we have eaten up the error.

For now let's throw the wrapped error (see comment in metricsTracker about how to wrap) so the bot crashes if we cannot send the startup event.
Also we should reference the underlying error and say this is the startup error in the wrapped message

cmd/trade.go
```diff
@@ -500,6 +505,21 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
+	tracker := metrics.MakeMetricsTracker(
```
nikhilsaraf, line 508: let's call this `metricsTracker`

cmd/trade.go
```diff
@@ -860,6 +885,12 @@ func deleteAllOffersAndExit(
 			return
 		}
 
+		e = tracker.SendDeleteEvent(l, true)
+		if e != nil {
+			l.Info(fmt.Sprintf("Sending deletion data event failed with error: %v\n", e))
```
nikhilsaraf, line 890: wrap and throw error.
can you use the word `metric` in all the log messages you've added? (example: delete event metric)
that will be very helpful in sifting through logs for these events

scripts/build.sh
```diff
@@ -235,6 +235,12 @@ then
     fi
 fi
 
+# TODO: Add check for -d (but not df).
+if [ -z "$AMPLITUDE_API_KEY" ]
```
nikhilsaraf, line 239: this should go on line 214 above.
It can check for non-force release like so: `if [[ FORCE_RELEASE -eq 0 ]]` see examples in the if-block near line 214 above.

also, in the same way that you've used `$AMPLITUDE_API_KEY` to access the env var here, you should inject it in the `LDFLAGS_ARRAY` array above. This can be set on the root.go file for now. See `version` variable as an example throughout the codebase.

Additionally, we need the kelp code to also do a check where if it is in release mode and does not have an API key it should crash the bot. this can be done in `runTradeCmd`

trader/trader.go
```diff
@@ -44,6 +46,7 @@ type Trader struct {
 	fixedIterations                *uint64
 	dataKey                        *model.BotKey
 	alert                          api.Alert
+	tracker                        *metrics.Tracker
```
nikhilsaraf, line 49: metricsTracker

trader/trader.go
```diff
@@ -117,6 +122,11 @@ func (t *Trader) Start() {
 		currentUpdateTime := time.Now()
 		if lastUpdateTime.IsZero() || t.timeController.ShouldUpdate(lastUpdateTime, currentUpdateTime) {
 			success := t.update()
+			if success {
+				l := logger.MakeBasicLogger()
+				go t.tracker.SendUpdateEvent(l)
```
nikhilsaraf, line 127: We had discussed using the `threadTracker` workflow to kick off a goroutine here so it is "managed". Unless there's a good reason to use an ad-hoc goroutine in this case, I think we should use the `threadTracker` approach.

nikhilsaraf: We should also send this once every X seconds (which can be a hard-coded number set to 600 in `metricTrackingUpdateIntervalSeconds` for now). This calculation is done by diffing against the `lastUpdateTime`. With that in mind, the `SendUpdateEvent` function should take in the `currentUpdateTime` for use in it's event.

We will need a new module that returns `metricTrackingUpdateIntervalSeconds` that takes in the startTime like we had discussed (within the first 2 hours it returns a value of 15 minutes, and after that it returns a value of 1 hour)

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+// SendStartupEvent sends the startup Amplitude event.
+func (t *Tracker) SendStartupEvent(l logger.Logger) error {
+	return t.sendEvent(l, "ce:test_startup", t.props)
+}
+
+// SendUpdateEvent sends the update Amplitude event.
+func (t *Tracker) SendUpdateEvent(l logger.Logger) error {
+	props := updateProps{
+		props:             t.props,
+		SecondsSinceStart: time.Now().Sub(t.start).Seconds(),
```
nikhilsaraf, line 116: should take in `time` as an argument and then compute the secondsSinceStart on it (see comment in `trader.go` for details)

nikhilsaraf (changes requested):
This PR has a fair amount of differences from what we had discussed. I've commented the specifics heavily to make this easier.

Answering questions above:
1. echo "$AMPLITUDE_API_KEY" should allow you to access the key, which you've done in the build.sh script. Echo also needs the `$` -- but we actually don't need echo at all. You can just use `"$AMPLITUDE_API_KEY"` directly and it should do the trick.

2. for the deletion event I'd recommend manually calling both events in the `update()` function to see if it sends the event. This should happen for both with exit and without

3. the function is is `deleteAllOffers` in trader.go. since it's a function on the trader object itself we don't need to thread the metricsTracker everywhere and can make a 1-line change by calling `t.SendDeleteEvent` directly

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+	"time"
+
+	"github.com/stellar/kelp/support/logger"
+
+	"github.com/google/uuid"
+)
+
+var (
+	amplitudeAPIURL string = "https://api2.amplitude.com/2/httpapi"
+	amplitudeAPIKey string = os.Getenv("AMPLITUDE_API_KEY")
```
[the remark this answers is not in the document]
debnil: Yup - the reason it isn't is because of the issue I outlined in 1 - wanted to commit something functional.

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+// TODO: Add cloud server information.
+type props struct {
+	BuildVersion string    `json:"build_version"`
+	Os           string    `json:"os"`
+	Gui          bool      `json:"gui"`
+	Strategy     string    `json:"strategy"`
+	UpdateTime   float64   `json:"update_time"`
+	Exchange     string    `json:"exchange"`
+	Pair         string    `json:"pair"`
+	SessionID    uuid.UUID `json:"session_id"`
```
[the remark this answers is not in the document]
debnil: SDKs would, but the HTTP API doesn't. API calls have no knowledge of the requests' session. The default value would be -1. [Relevant documentation](https://help.amplitude.com/hc/en-us/articles/115002323627-Tracking-Sessions#h_e9e9580e-5748-49ed-8ec4-8583c700ddd7).

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+
+// updateProps holds the properties for the update Amplitude event.
+type updateProps struct {
+	props
+	SecondsSinceStart float64 `json:"seconds_since_start"`
+}
+
+// deleteProps holds the properties for the delete Amplitude event.
+type deleteProps struct {
+	props
```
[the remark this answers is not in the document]
debnil: Right - the thinking here was that the specific properties structs would embed a struct that contains the "common" fields between all of them. Embedding different structs could create a more complex sense of dependency. This also allows greater modularity, in case we decide to add information to `updateProps` that we don't want in `deleteProps`.

That being said, the above involves a good deal of future proofing, and I'm happy to make this change now. Just flagging this thinking for future changes.

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// props holds the properties that we need for all Amplitude events.
+// This lives on the `Tracker` struct, although
+// TODO: Add geodata.
+// TODO: Add cloud server information.
+type props struct {
+	BuildVersion string    `json:"build_version"`
+	Os           string    `json:"os"`
+	Gui          bool      `json:"gui"`
```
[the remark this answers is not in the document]
debnil: Not sure I follow. The spec says that the `gui` data represents "triggered from GUI or not (see --ui flag)". This doesn't reference the GUI version - just how the user is using Kelp.

Should I use the `guiVersion`, rather than `options.gui`?

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+	Strategy     string    `json:"strategy"`
+	UpdateTime   float64   `json:"update_time"`
```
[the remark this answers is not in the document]
debnil: Got it. The confusion here was that the spec says "bot's update time interval in milliseconds", while the referenced field is denominated in seconds. Should this be multiplied by 1000, to fit the spec? Or is the original field name misleading?

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+
+	t := Tracker{
+		client: client,
+		props:  props,
+		start:  time.Now(),
+	}
+	return &t
+}
+
+func mustSessionID() uuid.UUID {
```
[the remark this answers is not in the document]
debnil: The `uuid` package includes the appropriate [function](https://pkg.go.dev/github.com/google/uuid#Must). I called it this way because no other `make*` factory method returns an error in the codebase. On a non-nil error, the `make*` functions in `cmd/trade.go` call `deleteAllOffersAndExit`, which is not exported or visible here.

As per this suggestion, I've bubbled up this error to the factory method, and in turn, called `deleteOffersAndExit` if the client creation fails. But two questions:

1. I thought we didn't want to fail if on `metricsTracker` failure. This change makes it impossible to run Kelp when compiling from source, unless either the end user provides an Amplitude API key or we distribute our API key to users.
2. A more general question: within the Kelp codebase, should factory methods return an error? I don't see this anywhere and am not sure of the general best practice in Go.

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// SendUpdateEvent sends the update Amplitude event.
+func (t *Tracker) SendUpdateEvent(l logger.Logger) error {
+	props := updateProps{
+		props:             t.props,
+		SecondsSinceStart: time.Now().Sub(t.start).Seconds(),
+	}
+
+	return t.sendEvent(l, "ce:test_update", props)
```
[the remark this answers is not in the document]
debnil: It's the best practice for naming custom events in Amplitude. In the web dashboard, it flags those events with the prefix `[Custom]`. 

Requested change made.

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+		SecondsSinceStart: time.Now().Sub(t.start).Seconds(),
+		Exit:              exit,
+		StackTrace:        "", // TODO: Determine how to do this.
+	}
+
+	return t.sendEvent(l, "ce:test_delete", props)
+}
+
+func (t *Tracker) sendEvent(l logger.Logger, eventType string, eventProps interface{}) error {
+	requestBody, e := json.Marshal(map[string]interface{}{
```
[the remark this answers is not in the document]
debnil: Argh, I meant to ask this in my initial questions, my bad. That method only works for request body of type `map[string]string`, while the serializable struct contains fields with non-string values. 

The method for type `map[string]interface{}`, `JSONRequestDynamicHeaders` makes sense on its own. But it requires a map of `HeaderFn` defined per-field. I tried to understand how to create this map from usage, but this function is only used in `support/sdk/ccxt`. The usage there is entirely contextual to the package and basically impossible to model. I could not figure out how to generate a comparable map for my own use case.

How do I do this? Do you have a helpful link or tutorial?

cmd/trade.go
debnil: Sounds good. Reasoning here is because the Go best practice is to avoid stuttering (my linter was heavily complaining with this name), but I get the need to align the codebase.

trader/trader.go
debnil: Ah, I definitely remember discussing the `threadTracker` usage - my bad. 

I don't think we discussed any new module or this differential behavior between the first 2 hours and later. Why do we want this approach?

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+	Props     interface{} `json:"event_properties"`
+}
+
+// props holds the properties that we need for all Amplitude events.
+// This lives on the `Tracker` struct, although
+// TODO: Add geodata.
+// TODO: Add cloud server information.
+type props struct {
+	BuildVersion string    `json:"build_version"`
+	Os           string    `json:"os"`
```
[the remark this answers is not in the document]
debnil: How would I define `goarm`? It's not a constant in the `runtime` package, and I don't think it's exported from the `build.sh` script.

debnil:
@nikhilsaraf: Thanks for the comprehensive and helpful review! I've covered the vast majority of it and made the relevant fixes. Some things still don't add up for me, and I've included these below.

1. How do I inject `goarm` to `tracker.props` `commonProps` struct? ([comment](https://github.com/stellar/kelp/pull/513#discussion_r495321678))
2. Using the `support/networking` package, as of now, is basically impossible without much more understanding of the networking header functions. Could you point me towards helpful resources? ([comment](https://github.com/stellar/kelp/pull/513#discussion_r495127618))
3. I can't totally parse the recommended checks for `build.sh`, as written out. Let me know if the changes I've made are the recommended ones. If not, a code sample would be really useful - it's difficult to convert Bash to English! :)
Related: within `runTradeCmd`, how do I check if currently in release mode? I've added a TODO for this, but wasn't sure how. ([comment](https://github.com/stellar/kelp/pull/513#discussion_r495022215))
4. The recommended time check for the `update` within `trader/trader.go` doesn't make much sense to me as written. Why would this require a new module? ([comment](https://github.com/stellar/kelp/pull/513#discussion_r495027363))

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: Ignore `JSONRequestDynamicHeaders()` for now. We need to use the `JSONRequest()` method. 

The `map[string]string` is for the `headers` variable.
It takes in the `data` field as a `string`. You will still need to marshall your data into a string like you've done here and then instead of using `http.Post` directly you should use `networking.JSONRequest()`

This function is pretty straightforward actually and not much different from `http.Post`. See callers of that function for usage examples. Most example use the `HTTP GET` but in our case we would use `"POST"` instead and instead of using an empty string for the `data` field it will be `string(requestBody)`, that's pretty much it, will make more sense once you find the usage examples in the codebase.

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: I think I added another comment at the point where this is used, to indicate how to inject in the guiVersion.
Basically you pass in `guiVersion` if `*options.gui` is true, otherwise you pass in `""`.

Yeah, you're right, the spec says it should be a boolean. It would be more useful to have the gui version as well -- we can have both if you think that would be useful, I figured we could achieve the same result by having the `guiVersion` as an empty string for whenever it wasn't triggered by the gui. But it doesn't cost us anything to also include the boolean flag so I'm fine with adding that if you think that will be easier for the query in amplitude.

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: yeah, looks like that's not available in the `runtime` package -- we can inject that in from the build script too, similar to how we inject the `version` string etc via LDFLAGS. then we will have a reference to it in `root.go`.

does that make sense?

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: yeah, the spec says milliseconds although it functions as seconds.
The initial thought was to be as fine-grained as possible but I think it's overkill. updated the spec to say seconds.
If we ever want finer-grained numbers we can add it as a separate field, so it should be good for now as seconds.

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: ah I see, that makes sense and I agree with your thinking and that's how I prefer to do it as well.
I was confused because we had used `props` in the initial event so thought the dependency grew out from there (so if we ever wanted to update the props sent to the initial event it would be sent everywhere else too which is bad).
Anyhow, with the suggestion to use `SecondsSinceStart` everywhere and rename to `commonProps` or similar would help disambiguate this. It would also be helpful to add a comment next to the struct definition with the thinking you described for posterity.

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: 1. yes, we need to have an alternative implementation to the metricTracker along with an interface etc., that has passthrough functions. We don't need this right now, and for now we can have a separate value passed in during construction called `enabled` which, as you might think, will not send the init and sendEvent API calls if it is disabled.

2. we have quite a few factory methods that return errors, you can find them using this regex `[Mm]ake[a-zA-Z]+\(.+\)\s+\(.*error\)`. This is not a go paradigm as much as it is a general programming pattern. Throwing/returning errors during constructions helps you to fail fast and therefore fix things that are broken a lot quicker.

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
nikhilsaraf: ok, that makes sense, can you add this as a comment next to the event names?

As a general note, comments in code are very useful to answer the "why" of things, although most programmers focus on comments addressing the "what" which is not super helpful. This is a perfect example of a comment that would help explain the why behind the "ce:" prefix.

cmd/trade.go
nikhilsaraf: yeah, there's a lot of fixes we need to make with the codebase to make them more go-idiomatic. But I'm less concerned about that because we have a fair amount of standardization as it is and that makes it easy(-ier) to make such changes.

trader/trader.go
nikhilsaraf: I didn't go into too much detail into the differential behavior because it's a nice to have and I didn't want us to lose focus from delivering the change. We can always add that as a separate commit/PR.
I'm fine with either including here or a separate PR, whatever you prefer.

And yes, a function works too.

support/metrics/metricsTracker.go
nikhilsaraf: when making this change, can you call the `time.Now()` variable as `now`? I started that pattern recently and want to standardize across the codebase (similar to how we have `e` for errors across the codebase)

nikhilsaraf (commented):
I'm not sure if I've addressed all the questions, but if things have been left out let's discuss today during our 1-1 or after that. Will be easier than over a crowded PR.

support/metrics/metricsTracker.go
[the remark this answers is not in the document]
debnil: TODO: File this as a follow-up issue, as per offline discussion. That issue also includes displaying it in `cmd/version`.

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return t.sendEvent(l, "ce:test_delete", props)
+}
+
+func (t *Tracker) sendEvent(l logger.Logger, eventType string, eventProps interface{}) error {
+	requestBody, e := json.Marshal(map[string]interface{}{
+		"api_key": amplitudeAPIKey,
+		"events": []event{event{
+			UserID:    "12345", // TODO: Determine actual user id.
```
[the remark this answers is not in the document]
debnil: TODO: File a separate issue for generating the user ID.

nikhilsaraf:
added an issue for the userId and the prefs file that we were discussing in the comments here: https://github.com/stellar/kelp/issues/516

cmd/trade.go
```diff
@@ -331,14 +333,15 @@ func makeStrategy(
 	options inputs,
 	threadTracker *multithreading.ThreadTracker,
 	db *sql.DB,
+	tracker *metrics.MetricsTracker,
```
nikhilsaraf, line 336: can we call this variable `metricsTracker` -- I thought we had changed this yesterday.

cmd/trade.go
```diff
@@ -496,10 +501,44 @@ func convertDeprecatedBotConfigValues(l logger.Logger, botConfig trader.BotConfi
 
 func runTradeCmd(options inputs) {
 	l := logger.MakeBasicLogger()
-	botConfig := readBotConfig(l, options)
+	botStart := time.Now()
+	botConfig := readBotConfig(l, options, botStart)
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
+	// TODO DS Fail if in release mode with undefined API key.
```
nikhilsaraf, line 509: this TODO no longer applies because we already added it in root.go

cmd/trade.go
```diff
@@ -496,10 +501,44 @@ func convertDeprecatedBotConfigValues(l logger.Logger, botConfig trader.BotConfi
...
+
+	userID := "-1" // TODO DS Properly generate and save user ID.
```
nikhilsaraf, line 511: see issue filed for this #516

cmd/trade.go
```diff
@@ -496,10 +501,44 @@ func convertDeprecatedBotConfigValues(l logger.Logger, botConfig trader.BotConfi
...
+
+	metricsTracker, e := metrics.MakeMetricsTracker(
+		userID,
+		amplitudeAPIKey,
+		httpClient,
+		botStart,
+		version,
+		runtime.GOOS,
+		runtime.GOARCH,
+		"", // TODO DS Determine how to get GOARM.
```
nikhilsaraf, line 526: `""` is a valid value for GOARM in the future (when it's missing).
so since it's unknown at this point, can you call it "unknown" so we don't incorrectly assign it as the valid value `""`

nikhilsaraf: alternatively we can choose a different value to represent a non-arm system such as `none`, your call but something to keep in mind for when this is addressed in the future

scripts/build.sh
```diff
@@ -188,7 +188,8 @@ GIT_BRANCH=$(git branch | grep \* | cut -d' ' -f2)
 VERSION_STRING="$GIT_BRANCH:$VERSION"
 GIT_HASH=$(git describe --always --abbrev=50 --dirty --long)
 DATE=$(date -u +%"Y%m%dT%H%M%SZ")
-LDFLAGS_ARRAY=("github.com/stellar/kelp/cmd.version=$VERSION_STRING" "github.com/stellar/kelp/cmd.guiVersion=$GUI_VERSION" "github.com/stellar/kelp/cmd.gitBranch=$GIT_BRANCH" "github.com/stellar/kelp/cmd.gitHash=$GIT_HASH" "github.com/stellar/kelp/cmd.buildDate=$DATE" "github.com/stellar/kelp/cmd.env=$ENV")
+echo "amplitude API KEY: $AMPLITUDE_API_KEY"
```
nikhilsaraf, line 191: we don't actually want to echo this here since it's included in the LDFLAGS_ARRAY
IIRC the echo was added during development to debug and should be part of the main repo?

scripts/build.sh
```diff
@@ -211,6 +212,16 @@ then
 
     if [[ IS_TEST_MODE -eq 0 ]]
     then
+        if [ -z "$AMPLITUDE_API_KEY" ]
```
nikhilsaraf, line 215: have you tested this by trying to do a release with the `-d` flag and the `-f` flag?

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,179 @@
...
+	"runtime/debug"
+	"time"
+
+	"github.com/stellar/kelp/support/networking"
+
+	"github.com/google/uuid"
+)
+
+var (
+	amplitudeAPIURL string = "https://api2.amplitude.com/2/httpapi"
```
nikhilsaraf, line 17: this could be a `const` instead of a var. non-blocking comment

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,179 @@
...
+	userID string
+	props  commonProps
+	start  time.Time
+}
+
+type event struct {
+	UserID    string      `json:"user_id"`
+	DeviceID  string      `json:"device_id"`
+	EventType string      `json:"event_type"`
+	Props     interface{} `json:"event_properties"`
```
nikhilsaraf, line 43: maybe as a separate PR once this is merged into master, can you see what additional properties (user properties and event properties) we can add on here to represent the official fields for things like OS, Country and other supported fields from Amplitude

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,179 @@
...
+	deleteProps := deleteProps{
+		commonProps: commonProps,
+		Exit:        exit,
+		StackTrace:  string(debug.Stack()),
+	}
+
+	return mt.sendEvent(deleteEventName, deleteProps)
+}
+
+// TODO DS Re-implement using `networking/JSONRequestDynamicHeaders`.
```
nikhilsaraf, line 150: can resolve this TODO

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,179 @@
...
+		"events": []event{event{
+			UserID:    mt.userID,
+			DeviceID:  mt.userID,
+			EventType: eventType,
+			Props:     eventProps,
+		}},
+	})
+
+	if e != nil {
+		log.Print("could not marshal json request")
```
nikhilsaraf, line 163: can remove this log.Print line

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,179 @@
...
+
+	log.Printf("Sending request: %v\n", string(requestBody))
+
+	var responseData interface{}
+	e = networking.JSONRequest(mt.client, "POST", amplitudeAPIURL, string(requestBody), map[string]string{}, &responseData, "")
+	if e != nil {
+		return fmt.Errorf("could not post amplitude request: %s", e)
+	}
+
+	log.Printf("Got response: %v\n", responseData)
```
nikhilsaraf, line 175: @debnil There are a total of three log lines here:
1. sending request (line 169)
2. got response (line 175)
3. successfully send event (line 177)

Instead of logging all three, we should remove (1) and (2) and change (3) to do this:
```
if strings.Contains(string(responseData), "success") {
    log.Printf("Successfully sent event metric of type '%s'", eventType)
} else {
    log.Printf("Failed to send event metric of type '%s' (request=%s; response: %s)", eventType, string(requestBody), string(responseData))
}
```

This will reduce the overhead of logging in relation to these requests and will at the same time give us the relevant information when needed.

PS: in the current state we will log line 177 (success) even if the response data indicates a failure. this will be fixed by the above suggestion

trader/trader.go
```diff
@@ -180,6 +191,10 @@ func (t *Trader) deleteAllOffers(isAsync bool) {
 			log.Fatalf("%scontinuing to exit after showing error during submission of delete offer ops: %s", logPrefix, e)
 			return
 		}
+
+		t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
+			t.metricsTracker.SendDeleteEvent(false)
```
nikhilsaraf, line 196: we are eating up the error from the return value here, we should check the error and log like we do for the `SendUpdateEvent() call on line 125`

nikhilsaraf (commented):
we're almost there. added some comments, particularly one about logging the events. see the original GH issue for all remaining follow-on work items to be handled in separate PRs/issues. If I've forgotten to list anything there please add for tracking.

cmd/trade.go
debnil: Awesome! Keeping the TODO for organizational purpose.

cmd/trade.go
debnil: Good call! Called it `unknown_goarm` just in case `unknown` is also valid - this should clearly disambiguate it.

scripts/build.sh
debnil: Good catch, this is a leftover from our pair debugging. Removing.

scripts/build.sh
debnil: Yup! Both exhibited the desired behaviors.

support/metrics/metricsTracker.go
debnil: Great suggestion for space-saving, but the suggested implementation (and variants of it) don't work cleanly. `responseData` is of `type interface{}`, so converting it to a string needs a type assertion. I've slightly modified this to define `responseDataStr := fmt.Sprintf("%v", responseData)

cmd/trade.go
debnil: Note that having an ID of -1 results in the HTTP response below. We'll have to prioritize this issue for Amplitude to be functional.

```response: map[code:400 error:Invalid id provided: omit the user_id or device_id field if they are unknown events_with_invalid_ids:map[device_id:[0] user_id:[0]]]```

**Comment by debnil** [comment]
@nikhilsaraf outstanding comments are all resolved, so I think this is good for a final look! Thanks so much for shepherding this PR through!

File: cmd/trade.go (hunk 25 above)
Comment (nikhilsaraf) [reply cmd/trade.go:509]: no worries, this will be disabled when checked-in at first, until we are ready to turn it on

File: trader/trader.go [hunk 35]
```diff
@@ -117,6 +121,13 @@ func (t *Trader) Start() {
 		currentUpdateTime := time.Now()
 		if lastUpdateTime.IsZero() || t.timeController.ShouldUpdate(lastUpdateTime, currentUpdateTime) {
 			success := t.update()
+			t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
```
nikhilsaraf, line 124: TODO NS: TriggerGoroutine() returns an error, we should check that

trader/trader.go
```diff
@@ -180,6 +191,13 @@ func (t *Trader) deleteAllOffers(isAsync bool) {
 			log.Fatalf("%scontinuing to exit after showing error during submission of delete offer ops: %s", logPrefix, e)
 			return
 		}
+
+		t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
```
nikhilsaraf, line 195: TODO NS: TriggerGoroutine() returns an error, we should check that

trader/trader.go
nikhilsaraf: TODO NS: send time taken to run update function in the update event

trader/trader.go
```diff
@@ -180,6 +191,13 @@ func (t *Trader) deleteAllOffers(isAsync bool) {
...
+			e := t.metricsTracker.SendDeleteEvent(false)
+			if e != nil {
+				log.Printf("failed to send update event metric: %s", e)
```
nikhilsaraf, line 198: TODO NS: comment why we are not throwing an error here

trader/trader.go
```diff
@@ -117,6 +121,13 @@ func (t *Trader) Start() {
 		currentUpdateTime := time.Now()
 		if lastUpdateTime.IsZero() || t.timeController.ShouldUpdate(lastUpdateTime, currentUpdateTime) {
 			success := t.update()
+			t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
+				e := t.metricsTracker.SendUpdateEvent(currentUpdateTime, success)
+				if e != nil {
+					log.Printf("failed to send update event metric: %s", e)
```
nikhilsaraf, line 127: TODO NS: comment why we are not throwing an error here

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,177 @@
...
+	deleteProps := deleteProps{
+		commonProps: commonProps,
+		Exit:        exit,
+		StackTrace:  string(debug.Stack()),
+	}
+
+	return mt.sendEvent(deleteEventName, deleteProps)
+}
+
+func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) error {
```
nikhilsaraf, line 149: TODO NS: `if mt.apiKey == ""` --> `return`

support/metrics/metricsTracker.go
```diff
@@ -0,0 +1,177 @@
...
+	}, nil
+}
+
+// SendStartupEvent sends the startup Amplitude event.
+func (mt *MetricsTracker) SendStartupEvent() error {
+	return mt.sendEvent(startupEventName, mt.props)
+}
+
+// SendUpdateEvent sends the update Amplitude event.
+func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool) error {
```
nikhilsaraf, line 126: TODO NS: also take in `millisForUpdate` property and add to struct accordingly

support/metrics/metricsTracker.go
nikhilsaraf: ref #516

nikhilsaraf (commented):
some comments as TODOs to be included before this can be merged (along with a round of manual testing to ensure no regressions). I'll address these TODOs to reduce the overhead of this PR which is mostly complete.

debnil:
Thanks! To clarify: are these comments that I should add in the codebase before merging? Or will you do that?

support/metrics/metricsTracker.go
nikhilsaraf: moving to #516

nikhilsaraf:
@debnil thanks for checking about these! The commented items are the leftover tasks to merge the code. It would actually be better if you could address them, do you think you will be able to get to them today?

debnil:
@nikhilsaraf got it - yes, can do today :)

trader/trader.go
debnil: Confused - I think this conflicts with [this](https://github.com/stellar/kelp/pull/513#discussion_r500364496)?

Seems like we should at least log the error in `Start`; I'm throwing a `log.Fatalf` because that's the pattern used elsewhere in the file for errors when triggering the goroutine.

trader/trader.go
debnil: Again, I think this conflicts with [this](https://github.com/stellar/kelp/pull/513#discussion_r500364496). I'm throwing a `log.Fatalf` around the error now, since that's the pattern used for other errors in the `deleteAllOffers` function.

support/metrics/metricsTracker.go
debnil: Added a TODO for this on the `commonProps` struct, seems like it requires a bit more implementation work and has been added to a follow-up issue as is.

support/metrics/metricsTracker.go
```diff
@@ -147,6 +148,10 @@ func (mt *MetricsTracker) SendDeleteEvent(exit bool) error {
 }
 
 func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) error {
+	if mt.apiKey == "" {
+		return fmt.Errorf("could not send amplitude event of type %s due to undefined API key", eventType)
```
nikhilsaraf, line 152: @debnil this should return `nil` not an error.

this is the scenario where we build without the api key, i.e. by compiling from source. If you run this bot now by compiling from source, I'd imagine that the bot is unable to start up because of this error?

trader/trader.go
nikhilsaraf: we want to check the error produced, which has been added, but we want to figure out what to do with the error. Also there are two errors, once in the context of the outer code, the other in the context of the inner goroutine function. Not sure if that's the conflict you were referring to (the link seems broken for me so I'm not sure)

We do not want to do a `log.Fatalf` because the error stems from a failure to send metrics. Logging with a `log.Printf` is the right thing to do. This is because of it being a metric more than it being from a goroutine. We don't want to crash the bot on metric issues (except at startup temporarily, which we do for now).

trader/trader.go
nikhilsaraf: see above for reasoning about using Printf instead of Fatalf for errors stemming from metrics.

I noticed a bigger issue at line 190 which I didn't catch earlier. When we call `SubmitOps` we pass in a function that is executed asynchronously. This function exits on success (or failure) - line 190. This means that we may never reach the `SendDeleteEvent` call. For this reason alone, we should move this TriggerGoroutine call (line 198) above the call to `SubmitOps` on line 190 (immediately after we enter the `if len(dOps) > 0` case).

nikhilsaraf (commented):
almost there. should be good to go after this. thanks for your perseverance!

nikhilsaraf (dismissed):
Looks great, thank you! 🎉 
I'm going to run this locally to do some sanity checks before merging.

nikhilsaraf (approved):
@debnil I've added some small changes to this PR, we were not doing some things correctly (session_id, capturing responses, etc).

It would be useful for us to go over these changes together on a call, I think it will help with improving overall code quality and reliability for future work.

PR #520, opened 2020-10-01
Disallow invalid characters in bot name (closes #429)

### LINKED ISSUE #429: [2] Kelp GUI: disallow invalid characters in bot name
## Description 

When naming a bot with characters such as `<` and or `>` it attempts to create a file with that name and also remove that file when deleting which fails:

```
error running rm command for bot configs: error in bash command 'rm /Users/user/.kelp/bot_data/configs/usd_<>_xlm*' for namespace 'rm': (eWait=exit status 1, outputBytes=, eRead=<nil>)
```

We should disallow names that would be invalid in the validation function on the server, simple fix for now.

## Your Environment
Kelp GUI

debnil:
**What**
This PR disallows invalid characters in the bot name. The screenshot displays the fix, surfaced in the GUI. This closes #429.
Disallow invalid characters in bot name

**Why**
From #428: "When naming a bot with characters such as `<` and or` >` it attempts to create a file with that name and also remove that file when deleting which fails."

**Known limitations**
N/A

### DIFF
--- a/gui/backend/upsert_bot_config.go
+++ b/gui/backend/upsert_bot_config.go
@@ -144,6 +144,11 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
 		}
 	}
 
+	invalidChars := "<>"
+	if strings.Contains(req.Name, invalidChars) {
+		errResp.Name = fmt.Sprintf("invalid bot name: cannot contain the following characters: %s", invalidChars)
+		hasError = true
+	}
 	if req.TraderConfig.AssetCodeA == "" || len(req.TraderConfig.AssetCodeA) > 12 {
 		errResp.TraderConfig.AssetCodeA = "1 - 12 characters"
 		hasError = true
--- a/gui/backend/upsert_bot_config.go
+++ b/gui/backend/upsert_bot_config.go
@@ -39,6 +39,9 @@ type upsertBotConfigResponseErrors struct {
 	Fields upsertBotConfigRequest `json:"fields"`
 }
 
+// botNameRegex is defined on initialization through the `cmd` package.
+var botNameRegex *regexp.Regexp
+
 func makeUpsertError(fields upsertBotConfigRequest) *upsertBotConfigResponseErrors {
 	return &upsertBotConfigResponseErrors{
 		Error:  "There are some errors marked in red inline",
@@ -145,13 +148,8 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
 		}
 	}
 
-	isValid, e := isBotNameValid(req.Name)
-	if !isValid {
-		if e != nil {
-			errResp.Name = "unable to parse bot name: internal server error"
-		} else {
-			errResp.Name = "invalid bot name: can only contain letters, numbers, spaces, or -"
-		}
+	if !isBotNameValid(req.Name) {
+		errResp.Name = "invalid bot name: can only contain letters, numbers, spaces, or -"
 		hasError = true
 	}
 	if req.TraderConfig.AssetCodeA == "" || len(req.TraderConfig.AssetCodeA) > 12 {
@@ -188,13 +186,21 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
 	return nil
 }
 
-func isBotNameValid(botName string) (bool, error) {
-	re, err := regexp.Compile(`^[a-zA-Z0-9\- ]+$`)
-	if err != nil {
-		return false, fmt.Errorf("could not compile botname regex")
+func isBotNameValid(botName string) bool {
+	if botNameRegex == nil {
+		panic(fmt.Errorf("botname regex undefined at runtime"))
 	}
 
-	return re.MatchString(botName), nil
+	return botNameRegex.MatchString(botName)
+}
+
+// InitBotNameRegex initializes the regex for bot names.
+func InitBotNameRegex() (e error) {
+	botNameRegex, e = regexp.Compile(`^[a-zA-Z0-9\- ]+$`)
+	if e != nil {
+		return fmt.Errorf("could not compile botname regex: %s", e)
+	}
+	return nil
 }
 
 func hasNewLevel(levels []plugins.StaticLevel) bool {
--- a/gui/backend/upsert_bot_config.go
+++ b/gui/backend/upsert_bot_config.go
@@ -196,6 +196,10 @@ func isBotNameValid(botName string) bool {
 
 // InitBotNameRegex initializes the regex for bot names.
 func InitBotNameRegex() (e error) {
+	if botNameRegex != nil {
+		return fmt.Errorf("botname regex already defined at init")
+	}
+
 	botNameRegex, e = regexp.Compile(`^[a-zA-Z0-9\- ]+$`)
 	if e != nil {
 		return fmt.Errorf("could not compile botname regex: %s", e)
--- a/gui/backend/upsert_bot_config.go
+++ b/gui/backend/upsert_bot_config.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/stellar/go/clients/horizonclient"
@@ -144,9 +145,13 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
 		}
 	}
 
-	invalidChars := "<>"
-	if strings.Contains(req.Name, invalidChars) {
-		errResp.Name = fmt.Sprintf("invalid bot name: cannot contain the following characters: %s", invalidChars)
+	isValid, e := isBotNameValid(req.Name)
+	if !isValid {
+		if e != nil {
+			errResp.Name = "unable to parse bot name: internal server error"
+		} else {
+			errResp.Name = "invalid bot name: can only contain letters, numbers, spaces, or -"
+		}
 		hasError = true
 	}
 	if req.TraderConfig.AssetCodeA == "" || len(req.TraderConfig.AssetCodeA) > 12 {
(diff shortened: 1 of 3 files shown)


nikhilsaraf (commented):
Your change addresses the specific bug that is outlined, but the filed issue is a symptom of the problem.

The problem we are trying to fix is that we don't validate the name. There are other characters that would trigger similar issues which we want to prevent. Is there a more generic way in which we can address this?

nikhilsaraf:
We want to ensure that we can save and delete files with names that we consider "valid". How can we test this?

debnil:
Both comments make sense. To answer the first, there are two pretty straightforward more generic addressing methods. Either we have a regex with the form of allowed filenames, or we have a list of denied characters within filenames (which just extends the existing solution). But the issue hasn't adequately defined the criteria for a "valid" filename. Without this, any solution cannot ensure that we can save and delete files as desired in a more general fashion.

If we want a general solution oriented around filenames, this is far from the "simple fix" suggested in the issue. Validity of filename depends on the operating system and specific file system, since Windows has a different set of invalid characters than UNIX. Relatedly, `<` and `>` aren't actually invalid characters in Unix filenames - they need a `\` - so I'm not sure if following filename conventions is the right approach. It seems like a road to a lot of special casing or two to three unmaintainable regexes.

I'm in favor of just expanding the existing denylist to all characters listed [here](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words). I don't think any of these characters are needed in a bot name, and it sidesteps errors like that documented in the issue. While this isn't a more general approach, it does give us much more control, without special-casing by operating system.

To answer your second question: testing either approach above isn't too hard. This can be done either manually or through adding a specific function to process the bot name, and adding unit tests for that function.

nikhilsaraf:
You make some good points.

We don't have a definition for a valid filename in the issue and it's interpretation is open-ended. Our solution will define it. The definition for valid should be derived from what the operating system allows, but since we aim to be compatible with a broad set of OSes, we can be restrictive in our definition.

An allowlist seems most appropriate since that avoids usage of characters we do not anticipate. Additionally, it gives us (developers) more flexibility even though it is more restrictive for the user.

We should be careful about spaces and underscores. We should only allow spaces in the "botName" because we do a conversion (encoding) of spaces to underscores in the "filename" (and back when loading the bot). No other characters require conversion/encoding from botName to filename. A basic set of characters in the allowlist should do the trick.

-----

As for testing, a unit test for the extracted function sounds good. We should have an automated test that uses the [`toml.WriteFile`](https://github.com/stellar/kelp/blob/master/gui/backend/upsert_bot_config.go#L96) method or [`ioutil.WriteFile`](https://golang.org/pkg/io/ioutil/#WriteFile) method in a temp directory, with the provided filename under test (table-based tests). Best to test all positive characters and some negative ones. Positive cases only should write the file and ensure it works, and then cleanup.

debnil:
Awesome! One additional clarification, just to make sure we're on the same page: I'm interpreting allow list as we should specify all valid characters (i.e., alphanumeric and spaces) and guarantee that bot names only contain characters in that set. I'm only clarifying because this approach differs from both outlined, so just making sure.

In any case, that testing strategy sounds good to me.

gui/backend/upsert_bot_config.go
```diff
@@ -144,6 +145,15 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
 		}
 	}
 
+	isValid, e := isBotNameValid(req.Name)
+	if !isValid {
+		if e != nil {
+			errResp.Name = "unable to parse bot name: internal server error"
```
nikhilsaraf, line 151: this is not an acceptable error to show the user on the input value. There needs to be a better flow.
with the changes to how the regex compile should work (comment below) this is not a situation we have to worry about now thankfully.

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,70 @@
...
+			wantValid: true,
+		},
+		{
+			name:      "success dash",
+			botName:   "George-the-Friendly-Octopus",
+			wantValid: true,
+		},
+		{
+			name:      "failure <>",
+			botName:   "George<>the Friendly Octopus",
```
nikhilsaraf, line 37: we should also test some other important invalid characters such as `.`, `&`, `_`, `(`, `)`, `+`, `@`

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,70 @@
...
+			botName:   "George\\/the Friendly Octopus",
+			wantValid: false,
+		},
+	}
+
+	for _, k := range testCases {
+		t.Run(k.name, func(t *testing.T) {
+			// errors would be thrown by the `regex` package, not due to kelp logic
+			gotValid, gotErr := isBotNameValid(k.botName)
+			assert.NoError(t, gotErr)
```
nikhilsaraf, line 51: if this is false then we should return so we don't produce false negatives for the remainder of the test:
```
if !assert.NoError(t, gotErr) {
    return
}
```

nikhilsaraf: can also use `e` instead of `gotErr` for errors that are not under test, which is a standardization across the codebase for tests.

using got in other places such as values (gotValid below) or errors being tested is ok.

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,70 @@
...
+			assert.Equal(t, k.wantValid, gotValid)
+			// early return on unsupported bot names, so we do not write an invalid filename
```
nikhilsaraf, line 53: this comment is extremely helpful ! 🎉

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,70 @@
...
+			wantValid: false,
+		},
+	}
+
+	for _, k := range testCases {
+		t.Run(k.name, func(t *testing.T) {
+			// errors would be thrown by the `regex` package, not due to kelp logic
+			gotValid, gotErr := isBotNameValid(k.botName)
+			assert.NoError(t, gotErr)
+			assert.Equal(t, k.wantValid, gotValid)
```
nikhilsaraf, line 52: return if this equal assertion is false for the same reason as above

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,70 @@
...
+			// early return on unsupported bot names, so we do not write an invalid filename
+			if !k.wantValid {
+				return
+			}
+
+			// confirm that we can write and delete an empty file to the bot's filepath
+			filenamePair := model2.GetBotFilenames(k.botName, "buysell")
+			fileBase, e := kelpos.MakeOsPathBase()
+			filePath := fileBase.Join(filenamePair.Trader)
+			assert.NoError(t, e)
```
nikhilsaraf, line 62: `e` should always be checked after it is produced (move one line above).

nikhilsaraf: should return if this check fails, here and everywhere else in this file too

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,70 @@
...
+			assert.NoError(t, gotErr)
+			assert.Equal(t, k.wantValid, gotValid)
+			// early return on unsupported bot names, so we do not write an invalid filename
+			if !k.wantValid {
+				return
+			}
+
+			// confirm that we can write and delete an empty file to the bot's filepath
+			filenamePair := model2.GetBotFilenames(k.botName, "buysell")
+			fileBase, e := kelpos.MakeOsPathBase()
```
nikhilsaraf, line 60: we don't want to use `MakeOsPathBase()` here. This will write to the user's production directory!
we should instead use a temporary directory like [`os.TempDir`](https://golang.org/pkg/os/#TempDir).

nikhilsaraf (changes requested):
some changes, feel like we should be good to go once these are fixed

gui/backend/upsert_bot_config.go
```diff
@@ -178,6 +188,15 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
 	return nil
 }
 
+func isBotNameValid(botName string) (bool, error) {
+	re, err := regexp.Compile(`^[a-zA-Z0-9\- ]+$`)
```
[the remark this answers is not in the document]
debnil: Compiling the regex on startup, rather than per-request, makes sense. But I don't see any existing function called `Init()` - are you suggesting creating this function? I'm implementing the rest of this logic in `cmd/server_amd64.go`, as requested.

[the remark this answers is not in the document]
debnil: Also, I'm not sure what "this file" refers to. I'm assuming you mean `gui/backend/upsert_bot_config.go`, because storing this variable within `cmd/server_amd64.go` as a global in the `cmd` package would create an import cycle.

gui/backend/upsert_bot_config_test.go
debnil: Makes sense, and `gotErr` use has been removed, since `isBotNameValid` no longer returns an error.

gui/backend/upsert_bot_config.go
```diff
@@ -178,6 +186,23 @@ func (s *APIServer) validateConfigs(req upsertBotConfigRequest) *upsertBotConfig
...
+func isBotNameValid(botName string) bool {
+	if botNameRegex == nil {
+		panic(fmt.Errorf("botname regex undefined at runtime"))
+	}
+
+	return botNameRegex.MatchString(botName)
+}
+
+// InitBotNameRegex initializes the regex for bot names.
+func InitBotNameRegex() (e error) {
```
nikhilsaraf, line 198: for higher quality code we should prevent double initialization, i.e. this should throw an error if botNameRegex is non-nil at the beginning of this function.

nikhilsaraf (approved):
LGTM, thanks!

gui/backend/upsert_bot_config_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+			// early return on unsupported bot names, so we do not write an invalid filename
+			if !k.wantValid {
+				return
+			}
+
+			// confirm that we can write and delete an empty file to the bot's filepath
+			filenamePair := model2.GetBotFilenames(k.botName, "buysell")
+			fileBase := os.TempDir()
+			filePath := fmt.Sprintf("%s%s", fileBase, filenamePair.Trader)
```
nikhilsaraf, line 99: @debnil the test is failing because of this -- we need a `/` separator between the `fileBase` and the trader filename

debnil: Interesting, this worked locally but not on CI. I wonder why the result of `os.TempDir` was processed differently.

debnil:
@nikhilsaraf: if you're open to it, could we change repo settings so that upon approval, the code owner review isn't needed for subsequent commits? Will help velocity when working asynchronously.

nikhilsaraf (approved):
LGTM, thanks! 🚢

PR #521, opened 2020-10-08
[WIP] Create frontend metrics tracker.

debnil:
This PR creates a frontend metrics tracker.

### DIFF
--- a/gui/web/src/App.js
+++ b/gui/web/src/App.js
@@ -9,7 +9,7 @@ import NewBot from './components/screens/NewBot/NewBot';
 import version from './kelp-ops-api/version';
 import quit from './kelp-ops-api/quit';
 import Welcome from './components/molecules/Welcome/Welcome';
-import {makeMetricsTracker} from './utils/metricsTracker';
+import MetricsTracker from './utils/metricsTracker';
 
 let baseUrl = function () {
   let origin = window.location.origin
@@ -31,7 +31,7 @@ class App extends Component {
 
     // metricsTracker sends UI events to Amplitude
     this.objects = {
-      metricsTracker: makeMetricsTracker(),
+      metricsTracker: MetricsTracker(""), // TODO DS Read API key from environment, as in Go backend.
     }
 
     this.setVersion = this.setVersion.bind(this);
@@ -46,6 +46,7 @@ class App extends Component {
 
   componentDidMount() {
     this.setVersion()
+    console.log(this.objects.metricsTracker.sendEvent("ce:test_startup"));
   }
 
   setVersion() {
--- a/gui/web/src/App.js
+++ b/gui/web/src/App.js
@@ -93,12 +93,12 @@ class App extends Component {
 
     // fetch object type from errors
     let kelp_errors = this.state.kelp_errors;
-    
+
     if (!kelp_errors.hasOwnProperty(backendError.object_type)) {
       kelp_errors[backendError.object_type] = {};
     }
     let botErrors = kelp_errors[backendError.object_type];
-
+    
     if (!botErrors.hasOwnProperty(backendError.object_name)) {
       botErrors[backendError.object_name] = {};
     }
@@ -161,14 +161,14 @@ class App extends Component {
     
     // return as an array
     return Object.values(levelErrors);
-  } 
+  }
 
   removeError(object_type, object_name, level, errorID) {
     let kelp_errors = this.state.kelp_errors;
     let botErrors = kelp_errors[object_type];
     let namedError = botErrors[object_name];
     let levelErrors = namedError[level];
-
+    
     // delete entry for error
     delete levelErrors[errorID];
     // bubble up
@@ -195,7 +195,7 @@ class App extends Component {
       const isLastError = newState.active_error.index > newState.active_error.errorList.length - 1;
       if (isLastError) {
         newState.active_error.index = newState.active_error.errorList.length - 1;
-      }
+      } 
       // else leave index as-is since we just deleted the index and the new item will now be at the old index (delete in place)
     }
     // trigger state change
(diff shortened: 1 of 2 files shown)


debnil:
@nikhilsaraf: This is very much a WIP/draft PR - I'm just putting out what I have for knowledge purposes, at the end of my workday, to get comments. I've chosen to export a `class` rather than a `function`, as it seems like the right pattern for object-oriented JavaScript, rather than nesting Javascript functions.

Outstanding questions:

1. Will the variables in the build script be exposed and accessible as public variables in the frontend, as in the Go backend?
2. When would `componentDidMount` be called, i.e., how do I test that this Amplitude event is sent? I'm not seeing any `console.log` in my terminal when I run the binary.
3. What's your linter setting in VSCode? I'm not sure what all I need to do to avoid making these changes in files I'm touching.

gui/web/src/App.js
```diff
@@ -40,6 +46,7 @@ class App extends Component {
 
   componentDidMount() {
     this.setVersion()
+    console.log(this.objects.metricsTracker.sendEvent("ce:test_startup"));
```
nikhilsaraf, line 49: we can just send the event, no need to log to console unless for local dev debugging.

gui/web/src/App.js
```diff
@@ -232,21 +239,21 @@ class App extends Component {
       <div>
         <div>{banner}</div>
         <Router>
-          <Header version={this.state.version}/>
+          <Header version={this.state.version} />
           <Route exact path="/"
-            render={(props) => <Bots {...props} baseUrl={baseUrl} activeError={this.state.active_error} setActiveError={this.setActiveBotError} hideActiveError={this.hideActiveError} addError={this.addError} removeError={removeBotError} getErrors={getBotErrors}/>}
-            />
+            render={(props) => <Bots {...props} baseUrl={baseUrl} activeError={this.state.active_error} setActiveError={this.setActiveBotError} hideActiveError={this.hideActiveError} addError={this.addError} removeError={removeBotError} getErrors={getBotErrors} />}
```
nikhilsaraf, line 244: can you update your git settings to ignore whitespaces?
not sure if there's a standardized way to do this but this is a [good starting point](https://stackoverflow.com/questions/3515597/add-only-non-whitespace-changes)

gui/web/src/utils/metricsTracker.js
```diff
@@ -0,0 +1,27 @@
+class MetricsTracker {
+    constructor(apiKey) {
+        this.apiKey = apiKey;
+        this.apiUrl = "https://api2.amplitude.com/2/httpapi";
+
+        // TODO: Implement more functions.
+        this.sendEvent = this.sendEvent.bind(this);
```
nikhilsaraf, line 7: should add this here too: `this._asyncRequests = {};`
the convention is to initialize all objects in the constructor so we know what all the objects are as opposed to finding new objects in a random function. this becomes more important as a given class grows.

gui/web/src/utils/metricsTracker.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    sendEvent(event) {
+        var _this = this
+        this._asyncRequests["fetch"] = fetch(this.apiUrl, {
```
nikhilsaraf, line 12: we shouldn't use the `fetch` function directly. everything should be in the kelp-ops-api folder (see version.js as an example).

this structures the code well when trying to streamline and track where we make all our network requests.

nikhilsaraf: we also don't want all requests to use the tag `fetch` since we don't want the sending of two different requests to cancel out each other. we can call it `"sendEvent/" + event` so each event gets its own namespace.

does the purpose of _asyncRequests make sense? (we use this pattern quite heavily across the app)

gui/web/src/utils/metricsTracker.js
```diff
@@ -0,0 +1,27 @@
...
+            method: "POST",
+            eventName: event,
+        }).then(resp => {
+            if (!_this._asyncRequests["fetch"]) {
+                // if it has been deleted, we don't want to process the result
+                return
+            }
+            delete _this._asyncRequests["fetch"];
+
+            return resp.json();
```
nikhilsaraf, line 22: we don't want to return the response really for sending events, unless there's an error. In the case where there's an error we can log to the console (because these are metric errors and should not be shown to the user in the GUI). In all other cases when it succeeds we can just ignore it really.

gui/web/src/App.js
nikhilsaraf: we should add the necessary startup items here, like we do in the CLI. don't we need to code that up in metricsTracker.js too?

nikhilsaraf (commented):
General form and structure of this PR looks good. specific comments inline.

When you try and run this, are there any errors that pop up? does it send the event to amplitude correctly?

gui/web/src/utils/metricsTracker.js
nikhilsaraf: PS context about some code style: infrastructure variables for non-product related bookkeeping (engineering related) are prefixed with `_` like here. `_asyncRequests` is the only example I can think off in the frontend codebase.

gui/web/src/utils/metricsTracker.js
debnil: Agreed that we shouldn't use the `fetch` function directly, but do we want to use `kelp-ops-api` for external requests? All existing endpoints in `kelp-ops-api` are for routes for the Kelp API (i.e., `api/v1/stop` or `api/v1/optionsMetadata`). Mixing internal and external use cases seems like a fast track to confusing directory structure. Let me know if there are already external use cases in that structure and I just missed it; if not, might be good to consider best organization. Thoughts on the right way to proceed here?

And yes, the purpose of `_asyncRequests` makes total sense. If I'm understanding the namespacing correctly, I can definitely change the namespace to `"sendEvent/" + event` - by modifying the key in `this._asyncRequests`, yeah?

gui/web/src/App.js
debnil: Yes, we do - but shouldn't those be added to the constructor, as we do in the CLI, rather than per-request?

nikhilsaraf: I was thinking about this a little and it's actually better to have a very different pattern. Let's reuse the infrastructure we have on the backend for this. We should create a new endpoint on the backend called `/sendMetricEvent` which takes in a JSON object as data, which it can add to the event props. It will reuse all the common props (other than 1 bit to distinguish this is come from the UI admin page)

the frontend would not need any metric specific logic and would just have one sendEvent.js function call that it can use to send events.

This will streamline all our metrics through one backend endpoint and common infrastructure which will make it much much less painful if we ever decide to switch providers or even do something like adding decorators on event tracking. This also moves a lot of the logic to the backend which is preferable

gui/web/src/utils/metricsTracker.js
nikhilsaraf: yeah, I agree -- we can create a new folder called `kelp-external-api` and put it in there. although with the above comment that will be routed through an internal request now. but good call!

re namespace: yes, we need to modify the key on the _asyncRequests map and that's pretty much it

debnil:
Closing this PR, since we decided to take a different architectural approach moving much of this logic to the backend.

PR #523, opened 2020-10-09
[WIP] Send frontend metrics.

debnil:
This WIP PR hooks the CLI metrics tracker to the GUI backend.

### DIFF
--- a/cmd/server_amd64.go
+++ b/cmd/server_amd64.go
@@ -33,6 +33,7 @@ import (
 	"github.com/stellar/kelp/gui/backend"
 	"github.com/stellar/kelp/support/kelpos"
 	"github.com/stellar/kelp/support/logger"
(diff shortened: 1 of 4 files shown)


debnil:
@nikhilsaraf: The goal with this PR is to align on a strategy to use the existing `metrics.MetricsTracker` in the GUI. I think this approach works well. We have two factory methods -- one for CLI, one for GUI -- so that we can keep using the existing common properties struct `commonProps`.

Then, on that struct, we mark the fields that are in the CLI but not the GUI as `omitempty`. Since those aren't given values in the GUI backend, they won't be serialized when GUI events are called, as they are empty on the `metrics.MarketTracker` struct.

Does this approach make sense to you? If so, I'll move on to the routes implementation, but I want to make sure we align on this first.

Issue #516, opened 2020-09-29
[8] follow-on tasks for Kelp CLI data tracking


nikhilsaraf:
## Desired Behavior
I want to persist state of user preferences and userId in regards to data tracking

## Impact
The desired behavior will allow me to be consistent with user prefs

## Features
- [x] add `isTestnet` field on commonProps (can apply to stellar testnet or sandboxed centralized exchanges) (#530) 
- [x] get deviceID using this library: [machineID](https://github.com/denisbrodbeck/machineid) (#531) 
- [x] user_id should be a hash of the trader account_id on sdex. for centralized exchanges it can be a hash of the API_KEY (#534)
- [x] disable event tracking if `--no-headers` flag specified in cli ([code](https://github.com/stellar/kelp/blob/master/cmd/trade.go#L516)) (#532)
- [x] add support for official amplitude properties (OS, Country, and other user/event properties)
- [x] send time taken to run update function in the update event as `millisForUpdate` and update relevant struct (#533)
- [x] ~extract call to Amplitude sendEvent API in `support/sdk/amplitude.go` (similar to `support/sdk/ccxt.go`)~ (don't need to do now)
- [ ] function to send update event more frequently in first hour and then slow down after that

## Additional context
related to #478

PR #530, opened 2020-10-13
Add isTestnet to metric common props (part of #516)

debnil:
This PR adds an `isTestnet` boolean field to the `commonProps` `struct` used in tracking metrics. Part of #516.

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -527,6 +527,7 @@ func runTradeCmd(options inputs) {
 		botConfig.TickIntervalSeconds,
 		botConfig.TradingExchange,
 		botConfig.TradingPair(),
+		strings.Contains(botConfig.HorizonURL, "test"),
 	)
 	if e != nil {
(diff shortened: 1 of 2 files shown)


cmd/trade.go
```diff
@@ -527,6 +527,7 @@ func runTradeCmd(options inputs) {
 		botConfig.TickIntervalSeconds,
 		botConfig.TradingExchange,
 		botConfig.TradingPair(),
+		strings.Contains(botConfig.HorizonURL, "test"),
```
nikhilsaraf, line 530: this should be false if trading on centralized exchanges

nikhilsaraf (commented):
one comment, rest looks good.

nikhilsaraf (approved):
looks good 🚢 
thanks for keeping these reviews small, they're much simpler to review.

when you commit these, if there's a way to make the commits show up in the GH issue (probably by referencing the issue number with a hash in the commit message) that would be great for future reference :)

PR #531, opened 2020-10-13
Add device ID to metric event props (part of #516)

debnil:
This PR adds the device ID to the metric tracker's event properties. Part of #516.

Before: the device ID is the same as the user ID.
Before: device ID equals user ID

After: the device ID is library generated and distinct from the user ID.
After: device ID is machine generated

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/denisbrodbeck/machineid"
 	"github.com/nikhilsaraf/go-tools/multithreading"
 	"github.com/spf13/cobra"
 
@@ -513,8 +514,14 @@ func runTradeCmd(options inputs) {
 		guiVersionFlag = guiVersion
 	}
 
+	deviceID, e := machineid.ID()
+	if e != nil {
+		logger.Fatal(l, fmt.Errorf("could not generate machine id: %s", e))
+	}
+
 	metricsTracker, e := metrics.MakeMetricsTracker(
 		userID,
+		deviceID,
 		amplitudeAPIKey,
 		httpClient,
 		botStart,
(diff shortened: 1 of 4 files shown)


nikhilsaraf:
I think the test is failing because the glide.yaml and glide.lock file is missing. Could you add these?
Getting the CircleCI tasks to pass is part of a complete PR.

In general, the broader questions to think about before merging code are:

- does this solve the problem in the bug/feature? If not then is the code deactivated somehow (feature flag / no path to possible execution)? Are there other cases beyond what was specifically mentioned in the bug/feature issue?
- am I comfortable merging this PR to the main branch? (no debug comments or print statements, etc.)
- is this going to cause any regressions? (circleCI, untested features)
- can a new developer get an exact copy of the kelp codebase that I currently have? (dependencies)

Looks good besides this :+1:

nikhilsaraf (commented):
I think the test is failing because the glide.yaml and glide.lock file is missing. Could you add these?
Getting the CircleCI tasks to pass is part of a complete PR.

In general, the broader questions to think about before merging code are:

- does this solve the problem in the bug/feature? If not then is the code deactivated somehow (feature flag / no path to possible execution)? Are there other cases beyond what was specifically mentioned in the bug/feature issue?
- am I comfortable merging this PR to the main branch? (no debug comments or print statements, etc.)
- is this going to cause any regressions? (circleCI, untested features)
- can a new developer get an exact copy of the kelp codebase that I currently have? (dependencies)

Looks good besides this :+1:

debnil:
100%. Not sure why the `glide.*` files didn't get added when I committed this PR - just realized this and will add.

nikhilsaraf (dismissed):
super, thanks!

nikhilsaraf (approved):

PR #532, opened 2020-10-13
Disable metrics tracking with no headers (part of #516)

debnil:
This PR disables metrics tracking if the `no-headers` flag is set to `true` in the CLI. Part of #516.

I confirmed this by setting the `userID=12345`, defining the API key, and running `kelp trade` with `no-headers=true`. I still got the expected log line on startup:
```metric - not sending event metric of type 'ce:test_startup' because metrics are disabled```

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -527,6 +527,7 @@ func runTradeCmd(options inputs) {
 		botConfig.TickIntervalSeconds,
 		botConfig.TradingExchange,
 		botConfig.TradingPair(),
+		*options.noHeaders, // disable metrics if the CLI specified no headers
 	)
 	if e != nil {
 		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker: %s", e))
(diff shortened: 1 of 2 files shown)


nikhilsaraf (approved):
lgtm, thanks 🚢

nikhilsaraf:
@debnil I think there was a bug while merging the code, which is why the test is failing. can you look into it?

nikhilsaraf (commented):
@debnil I think this is the wrong fix. if you look at the Files Changed, it includes references to `isTestnet` which is not a required change to get the disabled metrics issue working.

do we need isTestnet on the metrics tracker, or only on common props?
may help to compare to the [isTestnet commit](https://github.com/stellar/kelp/commit/98b31b2c93e9de93d0fd20c91792478ee1710693) in the master branch for diffs.

debnil:
Argh, thanks for catching. I'm getting mixed up between all of these PRs out. It should only be on commonProps, fixing now.

nikhilsaraf (approved):
🚢

PR #533, opened 2020-10-14
Send update event execution duration with metrics (part of #516)

debnil:
This PR sends the duration of the update event along with the metrics' common prop. Part of #516.

After: Can see `millisForUpdate` in Amplitude Web UI for an event.
Screen Shot 2020-10-13 at 6 15 47 PM

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -506,7 +506,7 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	userID := "-1" // TODO DS Properly generate and save user ID.
+	userID := "12345" // TODO DS Properly generate and save user ID.
 	httpClient := &http.Client{}
 	var guiVersionFlag string
 	if *options.ui {
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -506,7 +506,7 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	userID := "12345" // TODO DS Properly generate and save user ID.
+	userID := "-1" // TODO DS Properly generate and save user ID.
 	httpClient := &http.Client{}
 	var guiVersionFlag string
 	if *options.ui {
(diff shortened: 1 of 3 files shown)


cmd/trade.go
```diff
@@ -506,7 +506,7 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	userID := "-1" // TODO DS Properly generate and save user ID.
+	userID := "12345" // TODO DS Properly generate and save user ID.
```
nikhilsaraf, line 509: this seems like debug code since we still want the userID to be -1 for now (disabled), can you remove so it doesn't go into `master` (affects local dev builds)?

trader/trader.go
```diff
@@ -120,9 +120,11 @@ func (t *Trader) Start() {
 	for {
 		currentUpdateTime := time.Now()
 		if lastUpdateTime.IsZero() || t.timeController.ShouldUpdate(lastUpdateTime, currentUpdateTime) {
+			startUpdate := time.Now()
 			success := t.update()
+			millisForUpdate := time.Since(startUpdate).Milliseconds()
```
nikhilsaraf, line 125: should reuse `currentUpdateTime` instead of creating a new `startUpdate` time. Idea is to use one synchronized startTime variable wherever possible.

nikhilsaraf (commented):
mostly looks good, inline comment

trader/trader.go
debnil: Got it. What's the reason behind storing both `millisForUpdate` and `commonProps.SecondsSinceStart`? There's basically just two levels of functional indirection that differentiate these timestamps: triggering the goroutine and calling `SendUpdateEvent`. Is there significant overhead there?

cmd/trade.go
debnil: Thanks for catching!

trader/trader.go
nikhilsaraf: millisForUpdate is the time taken for the bot compute the action it wants to perform next, i.e. running the `update()` function (excluding sleep times or time waiting for transaction threads to finish). This can give us very useful performance metrics. For example, v1.11.0 is 10% faster than v1.10.0, or systems running windows is 20% slower than linux, as hypothetical examples.

secondsSinceStart and the different with `currentUpdateTime` tells us how many seconds it has been since the bot first started. This is useful in knowing things like long term stability of the bot. For example, say the bot always crashes if it is run for more than 5 hours -- this is what we can gauge from this metric.

nikhilsaraf (approved):
🚢

trader/trader.go
debnil: Got it, that makes sense -- thanks!

PR #534, opened 2020-10-14
Use hashed trading account for metrics user ID (part of #516)

debnil:
This PR uses the hash of the bot's trading account as the user ID for metrics collection. If the user is using a centralized exchange, then it uses the hash of the Amplitude API key. Part of #516.

After: We see a different integer as the User ID from those previously used, on a build from this branch.
After: unique integer user id, from this branch

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -731,7 +731,8 @@ func getUserID(l logger.Logger, botConfig trader.BotConfig) (string, error) {
 	if botConfig.IsTradingSdex() {
 		userIDPrehash = botConfig.TradingAccount()
 	} else {
-		userIDPrehash = amplitudeAPIKey
+		exchangeAPIKeys := botConfig.ExchangeAPIKeys.ToExchangeAPIKeys()
+		userIDPrehash = exchangeAPIKeys[0].Key
 	}
 
 	// hash avoids exposing the user account or api key
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -506,19 +506,11 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	var userIDPrehash string
-	if botConfig.IsTradingSdex() {
-		userIDPrehash = botConfig.TradingAccount()
-	} else {
-		userIDPrehash = amplitudeAPIKey
-	}
-
-	userIDHashed, e := utils.HashString(userIDPrehash)
+	userID, e := getUserID(l, botConfig)
 	if e != nil {
-		logger.Fatal(l, fmt.Errorf("could not create user id: %s", e))
+		logger.Fatal(l, fmt.Errorf("could not get user id: %s", e))
 	}
 
-	userID := fmt.Sprint(userIDHashed)
 	httpClient := &http.Client{}
 	var guiVersionFlag string
 	if *options.ui {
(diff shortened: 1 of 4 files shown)


cmd/trade.go
```diff
@@ -506,7 +506,19 @@ func runTradeCmd(options inputs) {
 	botConfig = convertDeprecatedBotConfigValues(l, botConfig)
 	l.Infof("Trading %s:%s for %s:%s\n", botConfig.AssetCodeA, botConfig.IssuerA, botConfig.AssetCodeB, botConfig.IssuerB)
 
-	userID := "-1" // TODO DS Properly generate and save user ID.
+	var userIDPrehash string
+	if botConfig.IsTradingSdex() {
+		userIDPrehash = botConfig.TradingAccount()
+	} else {
+		userIDPrehash = amplitudeAPIKey
```
nikhilsaraf, line 513: this should use the hash of the API key for the centralized exchange.
By using the amplitudeAPIKey, all users on centralized exchanges will have the same userID, which is not what we want here.

cmd/trade.go
```diff
@@ -506,7 +506,19 @@ func runTradeCmd(options inputs) {
...
+	}
+
+	userIDHashed, e := utils.HashString(userIDPrehash)
```
nikhilsaraf, line 516: can you add a comment describing why we take the hash?
reason is so we don't expose the user's account or api_key.

cmd/trade.go
```diff
@@ -506,7 +506,19 @@ func runTradeCmd(options inputs) {
...
+	if e != nil {
+		logger.Fatal(l, fmt.Errorf("could not create user id: %s", e))
+	}
+
+	userID := fmt.Sprint(userIDHashed)
```
nikhilsaraf, line 521: the newly added code is very self-contained. It may be cleaner if we put it in a separate function call, what do you think?
```
userID, e := getUserID(botConfig)
if e != nil {
    logger.Fatal(...)
}
```

cmd/trade.go
```diff
@@ -722,6 +726,26 @@ func runTradeCmd(options inputs) {
...
+		userIDPrehash = botConfig.TradingAccount()
+	} else {
+		exchangeAPIKeys := botConfig.ExchangeAPIKeys.ToExchangeAPIKeys()
+		userIDPrehash = exchangeAPIKeys[0].Key
+	}
+
+	// hash avoids exposing the user account or api key
+	userIDHashed, e := utils.HashString(userIDPrehash)
+	if e != nil {
+		logger.Fatal(l, fmt.Errorf("could not create user id: %s", e))
```
nikhilsaraf, line 741: wherever we can return an error we should. it's better than deciding how to handle the error at these inner function levels which is an unexpected "side effect" of the function.

cmd/trade.go
```diff
@@ -722,6 +726,26 @@ func runTradeCmd(options inputs) {
 	bot.Start()
 }
 
+func getUserID(l logger.Logger, botConfig trader.BotConfig) (string, error) {
+	var userIDPrehash string
+	if botConfig.IsTradingSdex() {
+		userIDPrehash = botConfig.TradingAccount()
+	} else {
+		exchangeAPIKeys := botConfig.ExchangeAPIKeys.ToExchangeAPIKeys()
+		userIDPrehash = exchangeAPIKeys[0].Key
```
nikhilsaraf, line 735: can the exchangeAPIKeys be empty here? if so it would result in an error, may be good to check the length and return an error if it is empty

cmd/trade.go
```diff
@@ -722,6 +726,26 @@ func runTradeCmd(options inputs) {
...
+
+	// hash avoids exposing the user account or api key
+	userIDHashed, e := utils.HashString(userIDPrehash)
+	if e != nil {
+		logger.Fatal(l, fmt.Errorf("could not create user id: %s", e))
+		return "", e
+	}
+
+	userID := fmt.Sprint(userIDHashed)
+	return userID, e
```
nikhilsaraf, line 746: error will always be nil here. can `return fmt.Sprint(userIDHashed), nil`
we do this as a convention across the codebase which gives the reader of the code confidence in knowing that all errors have been handled and bubbled up appropriately as soon as they were encountered. It makes the code a little easier to reason about as the codebase gets bigger

nikhilsaraf (commented):
thanks for extracting out the function. some comments about it inline

nikhilsaraf (approved):
🚢

PR #535, opened 2020-10-14
Add version to metric event.

debnil:
This PR adds the `version` to the top-level metric event. Note that based on the [new API docs](https://help.amplitude.com/hc/en-us/articles/360032842391-HTTP-API-V2#tocSsuccesssummary), this requires serializing the `app_version` as the field in JSON. 

This shows the `version` surfaced in the web console:
app_version shows up as Version


nikhilsaraf (approved):
🚢

PR #536, opened 2020-10-15
Send update metric at certain intervals (closes #516)

### LINKED ISSUE #516: [8] follow-on tasks for Kelp CLI data tracking

debnil:
This PR modifies our logic to sent the update metric, to send it at certain update intervals. Within 5 minutes of bot start, the interval is every 5 seconds; within 1 hour, every 10 minutes; and after that, every hour. Closes #516.


nikhilsaraf (commented):
some comments inline

trader/trader_test.go
```diff
@@ -159,3 +160,63 @@ func TestIsStateSynchronized(t *testing.T) {
...
+		start                time.Time
+		currentUpdate        time.Time
+		lastMetricUpdate     time.Time
+		wantShouldSendMetric bool
+	}{
+		{
+			name:                 "first 5 mins - refresh",
+			start:                now.Add(-4 * time.Minute),
+			currentUpdate:        now,
+			lastMetricUpdate:     now.Add(-10 * time.Second),
```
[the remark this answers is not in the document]
debnil: Tighter test condition makes sense. Happy to add the border test - my thinking was that this is just testing the math functionality in Go, rather than anything unique, but I can see it as a hedge against a typo.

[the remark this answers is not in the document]
nikhilsaraf: Yup, that’s mostly the benefit

trader/trader.go
```diff
@@ -115,21 +115,28 @@ func MakeTrader(
...
+					}
+				}, nil)
 				if e != nil {
-					log.Printf("failed to send update event metric: %s", e)
+					log.Printf("failed to trigger goroutine for send update event: %s", e)
 				}
-			}, nil)
-			if e != nil {
-				log.Printf("failed to trigger goroutine for send update event: %s", e)
+				lastMetricUpdateTime = currentUpdateTime
```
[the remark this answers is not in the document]
nikhilsaraf: Yeah, let’s add them, it’s cleaner that way and avoids any possible concurrency issue

trader/trader.go
```diff
@@ -107,6 +109,7 @@ func MakeTrader(
 		dataKey:                        dataKey,
 		alert:                          alert,
 		metricsTracker:                 metricsTracker,
+		start:                          start,
```
nikhilsaraf, line 112: In general I prefer calling variables like “startTime” instead of a more generic “start” to avoid having to refer to the type all the time for vars on structs, since they are not defined close to usage. For short scoped vars in functions it makes sense for shorter names like “start”.

nikhilsaraf: and as a convention input parameters on factory methods should have the same name as the field, overriding the short scope naming convention (camel cased if strict field is exported)

trader/trader_test.go
```diff
@@ -159,3 +160,63 @@ func TestIsStateSynchronized(t *testing.T) {
...
+	testCases := []struct {
+		name                 string
+		start                time.Time
+		currentUpdate        time.Time
+		lastMetricUpdate     time.Time
+		wantShouldSendMetric bool
+	}{
+		{
+			name:                 "first 5 mins - refresh",
+			start:                now.Add(-4 * time.Minute),
```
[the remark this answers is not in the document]
nikhilsaraf: @debnil can we set the start time to 5 minutes instead if 4?

Similarly for the hour boundaries— may uncover a <= bug

trader/trader.go
```diff
@@ -119,24 +122,23 @@ func (t *Trader) Start() {
...
 			success := t.update()
-			if shouldSendUpdateMetric(startTime, currentUpdateTime, lastMetricUpdateTime) {
+			if shouldSendUpdateMetric(t.start, currentUpdateTime, lastMetricUpdateTime) {
 				millisForUpdate := time.Since(currentUpdateTime).Milliseconds()
 				e := t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
 					e := t.metricsTracker.SendUpdateEvent(currentUpdateTime, success, millisForUpdate)
 					if e != nil {
 						log.Printf("failed to send update event metric: %s", e)
 					}
+					lastMetricUpdateTime = currentUpdateTime
```
nikhilsaraf, line 137: Not sure where my comment went — yes, let’s please add the last metric updated time for send update on the metrics tracker. Will avoid all potential concurrency issues

nikhilsaraf (commented):
Should be ready to merge after this round of reviews I think

trader/trader.go
```diff
@@ -121,15 +124,18 @@ func (t *Trader) Start() {
...
-			e := t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
-				e := t.metricsTracker.SendUpdateEvent(currentUpdateTime, success, millisForUpdate)
+			if shouldSendUpdateMetric(t.startTime, currentUpdateTime, t.metricsTracker.LastMetricUpdateTime) {
+				millisForUpdate := time.Since(currentUpdateTime).Milliseconds()
+				e := t.threadTracker.TriggerGoroutine(func(inputs []interface{}) {
+					e := t.metricsTracker.SendUpdateEvent(currentUpdateTime, success, millisForUpdate)
+					if e != nil {
+						log.Printf("failed to send update event metric: %s", e)
+					}
+					t.metricsTracker.LastMetricUpdateTime = currentUpdateTime
```
nikhilsaraf, line 134: I think I may not have communicated my intent very clearly here. My bad.

We don’t want to update the time stamp here, especially not as an exported variable. Instead, we want the metricsTracker to update it on itself in the sendUpdsteEvent function on success.

support/metrics/metricsTracker.go
```diff
@@ -24,13 +24,14 @@ const (
...
-	start      time.Time
-	isDisabled bool
+	client               *http.Client
+	apiKey               string
+	userID               string
+	deviceID             string
+	props                commonProps
+	botStartTime         time.Time
+	isDisabled           bool
+	LastMetricUpdateTime time.Time
```
nikhilsaraf, line 34: This should be unexported and should have an exported Get method. See second comment to get full pattern of usage.

Also, the name is very generic now that it’s on the struct. It’s not clear if this represents the last metric update time of the startup event, delete event, update event, or any event.

You and I know what it stands for because we have context on it. That context should be captured in the code preferably with the variable name. Maybe something like updateEventSentTime

nikhilsaraf (commented):
Comment about antipattern, rest looks good

nikhilsaraf (approved):
this looks great, thanks! 🎉

Issue #508, opened 2020-09-17
[13] Implement Metrics in the Kelp GUI


nikhilsaraf:
Follow what is described in the doc:
https://docs.google.com/document/d/1R5hU7dSloQHZ0XNZkUrS0LPOr2gLVpteXq5NvR-Fy2k/edit#

This is one of the last items needed to release Kelp GUI rc2

nikhilsaraf:
marked as a 21-pointer because of the number of items that need to be tracked. Can break down into smaller tasks based on the titles in the document, let's discuss in sprint planning when you get to this issue.

PR #549, opened 2020-10-20
Enable GUI metrics tracking (part of #508)

debnil:
This PR enables the use of the metricsTracker on the GUI. This includes:
* Separate factory methods for the CLI and GUI
* Adding the `metricsTracker` onto the `apiServer`
* Routes in both the GUI and backend for the `sendMetricEvent`


debnil:
@nikhilsaraf This PR sets up all the infrastructure to send Amplitude events from the UI to the backend. If I should also include an event in this PR, let me know.

cmd/server_amd64.go
```diff
@@ -288,6 +289,28 @@ func init() {
 			}
 		}
 
+		var metricsTracker *metrics.MetricsTracker
+		if isLocalDevMode {
+			log.Printf("not sending data metrics in dev mode")
```
nikhilsaraf, line 294: prefix log line with "metric - " for easy grep-ability in logs

cmd/server_amd64.go
```diff
@@ -288,6 +289,28 @@ func init() {
...
+		} else {
+			userID := "-1" // TODO DS Determine how to get user ID in GUI mode.
```
nikhilsaraf, line 296: oh yeah, this is another issue altogether! we can just use deviceID here since we don't have a concept of a user.

gui/backend/send_metric_event.go
```diff
@@ -0,0 +1,42 @@
...
+	log.Printf("sendMetricEvent requestJson: %s\n", string(bodyBytes))
+
+	var req sendMetricEventRequest
+	e = json.Unmarshal(bodyBytes, &req)
+	if e != nil {
+		s.writeErrorJson(w, fmt.Sprintf("error unmarshaling json: %s; bodyString = %s", e, string(bodyBytes)))
+		return
+	}
+
+	e = s.metricsTracker.SendGuiEvent(req.EventName)
```
nikhilsaraf, line 35: the function signature should also take in a `map[string]interface{}`. but if we don't need it in this PR then can expand the signature later

gui/backend/send_metric_event.go
```diff
@@ -0,0 +1,42 @@
+package backend
+
+import (
```
nikhilsaraf, line 3: lgtm

support/metrics/metricsTracker.go
```diff
@@ -56,10 +56,10 @@ type commonProps struct {
 	Goarch                    string  `json:"goarch"`
 	Goarm                     string  `json:"goarm"`
 	GuiVersion                string  `json:"gui_version"`
-	Strategy                  string  `json:"strategy"`
-	UpdateTimeIntervalSeconds int32   `json:"update_time_interval_seconds"`
-	Exchange                  string  `json:"exchange"`
-	TradingPair               string  `json:"trading_pair"`
+	Strategy                  string  `json:"strategy,omitempty"`
```
nikhilsaraf, line 59: have you tested `omitempty` to see if it works as expected?

support/metrics/metricsTracker.go
```diff
@@ -157,6 +186,20 @@ func (mt *MetricsTracker) SendStartupEvent() error {
 	return mt.sendEvent(startupEventName, mt.props)
 }
 
+// SendGuiEvent sends an event from the GUI.
+// TODO DS Determine generic properties to send with the GUI event.
+func (mt *MetricsTracker) SendGuiEvent(eventName string) error {
```
nikhilsaraf, line 191: when you have two factory methods -- makeCLi and makeGui and then two methods on it sendEvent and sendGuiEvent it's a sign that we need an interface and two separate implementations.

A good way to know this is if the developer needs to know that we cannot use sendEvent for gui based metrics -- that's a sign.

can we not reuse the `sendEvent` method and export that?

nikhilsaraf: if we have logic that we always send the `mt.props` and merge in any values passed in the `map[string]interface{}` into `mt.props` then we can achieve this -- i.e. the merging can happen in sendEvent as opposed to in `updateProps`

cmd/server_amd64.go
debnil: Sounds good! I'm passing the device ID for both the user ID and the device ID - seems like we might as well populate both, it'll make it easier to differentiate in the web UI too.

gui/backend/send_metric_event.go
debnil: Agreed - there's a TODO for this above `SendGuiEvent`'s definition in `metricsTracker.go`.

support/metrics/metricsTracker.go
debnil: How do you do this merge? I get what needs to be done, but I can't figure out a syntax that would work.

debnil: Never mind, I've figured out an approach that generates correct output. I don't think there's another way to merge two arbitrary structs, and it works for our purposes. Let me know what you think.

gui/backend/send_metric_event.go
debnil: Updated: This function no longer exists. Instead, `metricsTracker.SendEvent` already takes an arbitrary interface.

gui/web/src/kelp-ops-api/sendMetricEvent.js
```diff
@@ -0,0 +1,8 @@
+export default (baseUrl, eventData) => {
```
nikhilsaraf, line 1: would it be a better API if this took `name` and `eventData` as two explicit fields rather than combining into one as `eventData`?

`name` is required, but `eventData` is optional -- this is how the backend data works (`sendMetricEventRequest` struct) in `send_metric_event.go` and should be reflected in the frontend API function.

this could be formatted as the following json object:
```
{
    name: "someName",
    data: { ... }
}
```

gui/backend/send_metric_event.go
```diff
@@ -0,0 +1,42 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+type sendMetricEventRequest struct {
+	EventType  string      `json:"type"`
```
nikhilsaraf, line 12: can we call this `event_type` to be a little more specific?

gui/backend/send_metric_event.go
```diff
@@ -0,0 +1,42 @@
...
+	EventProps interface{} `json:"props"`
```
nikhilsaraf, line 13: and `event_props` -- helps to distinguish between `user_props`, since this will not be merged with user props right now

support/metrics/metricsTracker.go
```diff
@@ -190,34 +188,65 @@ func MakeMetricsTracker(
...
+	client *http.Client,
+	botStartTime time.Time,
+	version string,
+	goos string,
+	goarch string,
+	goarm string,
+	guiVersion string,
+) (*MetricsTracker, error) {
+	props := commonProps{
+		CliVersion: version,
```
nikhilsaraf, line 205: we may need to add some more fields from what was recently added such as `GitHash` and `Env`. Don't think there's any others that are applicable from the list.

support/metrics/metricsTracker.go
```diff
@@ -190,34 +188,65 @@ func MakeMetricsTracker(
...
+		GuiVersion: guiVersion,
+	}
+
+	return &MetricsTracker{
+		client:       client,
+		apiKey:       apiKey,
+		userID:       userID,
+		deviceID:     deviceID,
+		props:        props,
+		botStartTime: botStartTime,
```
nikhilsaraf, line 218: explicitly set all fields on the struct -- i.e. isDisabled and `updateEventSentTime`
should be taken in as an argument like we do for the cli

support/metrics/metricsTracker.go
```diff
@@ -190,34 +188,65 @@ func MakeMetricsTracker(
...
 func (mt *MetricsTracker) SendStartupEvent() error {
-	return mt.sendEvent(startupEventName, mt.props)
+	return mt.SendEvent(startupEventName, mt.props)
 }
 
 // SendUpdateEvent sends the update Amplitude event.
 func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisForUpdate int64) error {
-	commonProps := mt.props
-	commonProps.SecondsSinceStart = now.Sub(mt.botStartTime).Seconds()
+	mt.props.SecondsSinceStart = now.Sub(mt.botStartTime).Seconds()
```
nikhilsaraf, line 234: why are we setting this on the `mt.props` vs on a new copy of the props object? I feel like the previous way was better with `commonProps`.

(will need to check the golang semantics of where/when we create a copy of structs)

support/metrics/metricsTracker.go
```diff
@@ -228,33 +257,37 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
 
 // SendDeleteEvent sends the delete Amplitude event.
 func (mt *MetricsTracker) SendDeleteEvent(exit bool) error {
-	commonProps := mt.props
-	commonProps.SecondsSinceStart = time.Now().Sub(mt.botStartTime).Seconds()
+	mt.props.SecondsSinceStart = time.Now().Sub(mt.botStartTime).Seconds()
```
nikhilsaraf, line 260: same here -- mt.props vs commonProps -- whatever solution we decide to move forward with

support/metrics/metricsTracker.go
```diff
@@ -228,33 +257,37 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
...
+		Exit:       exit,
+		StackTrace: string(debug.Stack()),
 	}
 
-	return mt.sendEvent(deleteEventName, deleteProps)
+	return mt.SendEvent(deleteEventName, deleteProps)
 }
 
-func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) error {
+// SendEvent sends an event with its type and properties to
```
nikhilsaraf, line 269: this comment sentence seems incomplete.
let's finish the commend, or remove the "to"

support/metrics/metricsTracker.go
```diff
@@ -286,3 +319,28 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
 	}
 	return nil
 }
+
+func (mt *MetricsTracker) mergeEventProps(eventProps interface{}) (map[string]interface{}, error) {
```
nikhilsaraf, line 323: this function needs a test

if it's easier to write the test, this can be a function that's not on `metricsTracker` and is a standalone function. This can have the signature of `(parentProps map[string]interface{}, additionalProps map[string]interface{}) (map[string]interface{}, error)`

It may even be good to put this in `utils/functions.go` along with the test in `utils/functions_test.go` calling it `mergeMaps` or something. Your call on best approach.

support/metrics/metricsTracker.go
```diff
@@ -286,3 +319,28 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
...
+	var m map[string]interface{}
+	eventJSON, e := json.Marshal(eventProps)
+	if e != nil {
+		return nil, fmt.Errorf("metric - could not marshal event props json: %s", e)
+	}
+
+	e = json.Unmarshal(eventJSON, &m)
```
nikhilsaraf, line 330: marshalling and unmarshalling so many times seems excessive and expensive -- there must be a better way of converting from a `struct` to `map[string]interface{}`.

wonder if this library can help (we already depend on it in `ccxt.go`): [github.com/mitchellh/mapstructure](github.com/mitchellh/mapstructure)

nikhilsaraf (commented):
added comments

support/metrics/metricsTracker.go
nikhilsaraf: discussed on call, decided to move mt.props.SecondSinceStart into `SendEvent` because it's common to all events. Callers should pass in a `now` field.

support/metrics/metricsTracker.go
```diff
@@ -190,34 +188,65 @@ func MakeMetricsTracker(
...
+
 // GetUpdateEventSentTime gets the last sent time of the update event.
 func (mt *MetricsTracker) GetUpdateEventSentTime() *time.Time {
 	return mt.updateEventSentTime
 }
 
 // SendStartupEvent sends the startup Amplitude event.
 func (mt *MetricsTracker) SendStartupEvent() error {
-	return mt.sendEvent(startupEventName, mt.props)
+	return mt.SendEvent(startupEventName, mt.props)
```
nikhilsaraf, line 229: @debnil how we can convert SendEvent's second param to map[string]interface{}
```
v, e := toMapStringInterface(mt.props)
// check e
return mt.SendEvent(startupEventName, v)

func toMapStringInterface(v interface{}) (map[string]interface{}, error) {
        var m map[string]interface{}
	eventJSON, e := json.Marshal(v)
	if e != nil {
		return nil, fmt.Errorf("could not marshal to map[string]interface{}: %s", e)
	}
}
```

support/metrics/metricsTracker.go
```diff
@@ -190,34 +188,65 @@ func MakeMetricsTracker(
...
+		Goarm:      goarm,
+		GuiVersion: guiVersion,
+	}
+
+	return &MetricsTracker{
+		client:       client,
+		apiKey:       apiKey,
+		userID:       userID,
+		deviceID:     deviceID,
+		props:        props,
```
nikhilsaraf, line 217: we should set the field `props` to a `map[string]interface{}` on metricsTracker -- we do this once only using the `toMapStringInterface` function and then in mergeProps we can merge two maps instead of doing any json work

support/metrics/metricsTracker.go
debnil: I don't think this method quite works, since `json.Marshal` creates a `[]byte`. To create a `map[string]interface{}`, I would then have to call `json.Unmarshal` on the resulting byte-array - which is the exact approach we tried to avoid.

I think all we need to do is some syntactic sugar around a typecast to `map[string]interface{}`, instead, which I've added in `toMapStringInterface`. This checks the `ok` boolean and returns an error if false.

support/metrics/metricsTracker.go
debnil: Post-rewrite, is a unit test still worthwhile? Now, it just merges two maps into one - it iterates over each and assigns to a new map. I've written the test, but not sure it's still needed.

gui/web/src/kelp-ops-api/sendMetricEvent.js
```diff
@@ -0,0 +1,11 @@
+export default (baseUrl, name, eventData) => {
+    return fetch(baseUrl + "/api/v1/sendMetricEvent", {
+        method: "POST",
+        body: {
+            name: name,
```
nikhilsaraf, line 5: format is good, but shouldn't this be `event_name` and `event_data` as the fields?

this should have been caught in testing before submitting as a PR. it's always best practice to test your code before submitting PRs so we can avoid raising such issues

support/metrics/metricsTracker.go
```diff
@@ -286,3 +350,36 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
...
+	}
+
+	for k, v := range eventProps {
+		m[k] = v
+	}
+
+	return m, nil
+}
+
+func toMapStringInterface(v interface{}) (map[string]interface{}, error) {
```
nikhilsaraf, line 367: this seems like a generic enough function to where we can put this in `utils.functions.go` - along with a comment describing what it does since it deals with a lot of `interface{}` types so is hard to tell what it does without reading the code. quick one-line comment will help - can use an example to help if you want.

support/metrics/metricsTracker.go
```diff
@@ -286,3 +350,36 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
...
+	b, e := json.Marshal(v)
+	if e != nil {
+		return nil, fmt.Errorf("could not marshal interface to json: %s", e)
+	}
+
+	var i interface{}
```
nikhilsaraf, line 373: wondering if this can be a `map[string]interface{}`?

then we won't need the cast below

support/metrics/metricsTracker.go
```diff
@@ -286,3 +350,36 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
...
+	e = json.Unmarshal(b, &i)
+	if e != nil {
+		return nil, fmt.Errorf("could not unmarshal json to interface: %s", e)
+	}
+
+	m, ok := i.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("could not create map[string]interface{}")
```
nikhilsaraf, line 381: we should also specify the type here in the error message:
`fmt.Errorf("could not create map[string]interface{}, underlying type was '%T'", i)`

not applicable if using `map[string]interface{}` works above

support/metrics/metricsTracker_test.go
```diff
@@ -0,0 +1,42 @@
...
+		"4": "four",
+	}
+
+	gotMap, e := mergeEventProps(commonProps, eventProps)
+	assert.Equal(t, wantMap, gotMap)
+	assert.NoError(t, e)
+}
+
+func TestToMapStringInterface(t *testing.T) {
+	success := map[string]interface{}{
```
nikhilsaraf, line 31: this input is not the most helpful since we convert from a map to a map. while this is a useful case to have, it's more interesting to test whether a struct can be converted to a map, especially since that's our use case. we can add that as a third case.

these three cases are fairly distinct, and we don't anticipate having many tests for which table based tests would make sense. It's probably better to separate out these cases into three different tests or as three subtests in this one test by using t.Run("success_map", ...), "success_struct" and then "failure_float"

support/metrics/metricsTracker.go
```diff
@@ -196,28 +244,28 @@ func (mt *MetricsTracker) GetUpdateEventSentTime() *time.Time {
...
-	return mt.sendEvent(startupEventName, mt.props)
+func (mt *MetricsTracker) SendStartupEvent(now time.Time) error {
+	return mt.SendEvent(startupEventName, mt.props, now)
 }
 
 // SendUpdateEvent sends the update Amplitude event.
 func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisForUpdate int64) error {
-	commonProps := mt.props
-	commonProps.SecondsSinceStart = now.Sub(mt.botStartTime).Seconds()
+	secondsSinceStart := now.Sub(mt.botStartTime).Seconds()
```
nikhilsaraf, line 253: we don't need this here, can put into the if condition where updateEventSentTime is nil since we don't set it explicitly on the commonProps anymore

support/metrics/metricsTracker.go
```diff
@@ -228,34 +276,50 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
...
 		log.Printf("metric - not sending event metric of type '%s' because metrics are disabled", eventType)
 		return nil
 	}
 
+	trackerProps := mt.props
+	trackerProps[secondsSinceStartKey] = now.Sub(mt.botStartTime).Seconds()
+
+	cliVersion, ok := trackerProps[cliVersionKey].(string)
+	if !ok {
+		return fmt.Errorf("could not cast cli version back to string")
```
nikhilsaraf, line 299: it would be better to save the cliVersion on the `mt` directly rather than doing this check every time. they both achieve the same thing but it's cleaner this way and allows us to retain the type information about the cliVersion field (which is now lost since props is a map[string]interface{})

support/metrics/metricsTracker.go
```diff
@@ -286,3 +350,36 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
 	}
 	return nil
 }
+
+func mergeEventProps(commonProps, eventProps map[string]interface{}) (map[string]interface{}, error) {
```
nikhilsaraf, line 354: commented in another pull request to avoid the pattern of input vars like `(v1, v2 int)` and instead use the more explicit version of `(v1 int, v2 int)`

nikhilsaraf: this function should not contain context of its usage -- here we are holding onto the context in the variable names with commonProps and eventProps.

better names would be `original` and `overrides`

nikhilsaraf: this is also a generic enough function that it can be moved to `utils/functions.go`

support/metrics/metricsTracker.go
```diff
@@ -286,3 +350,36 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
...
+	m := make(map[string]interface{})
```
nikhilsaraf, line 355: any benefits to using `make` instead of `map[string]interface{}{}`?

support/metrics/metricsTracker_test.go
```diff
@@ -0,0 +1,42 @@
+package metrics
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMergeEventProps(t *testing.T) {
```
nikhilsaraf, line 9: 👍

nikhilsaraf (commented):
some comments inline, nothing major. should be good to merge in next iteration.

support/metrics/metricsTracker_test.go
nikhilsaraf: can you add the following types as values in the struct?:
bool, string, int, float, struct, []int

cmd/trade.go
```diff
@@ -603,7 +603,7 @@ func runTradeCmd(options inputs) {
 		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker: %s", e))
 	}
 
-	e = metricsTracker.SendStartupEvent()
+	e = metricsTracker.SendStartupEvent(time.Now())
```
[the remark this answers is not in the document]
debnil: Sure thing - I agree with this, this was just an oversight in prior PRs. Added now :)

gui/web/src/kelp-ops-api/sendMetricEvent.js
debnil: Thanks for catching, works now - I just missed a manual test with all the back and forth.

support/metrics/metricsTracker.go
debnil: I think it can be, but it's recommended everywhere I could find resources to unmarshal arbitrary structs to `map[string]interface{}`. See a relevant Golang [blogpost](https://blog.golang.org/json), under "Decoding arbitrary data".

With that said, I can't find a functional reason for this and it saves a cast that should get caught in the Unmarshal, so making this change.

support/metrics/metricsTracker.go
debnil: I think that throws a segfault upon the initial assignment, since the memory is not allocated.

support/metrics/metricsTracker.go
```diff
@@ -246,34 +295,45 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
...
+		APIKey: mt.apiKey,
 		Events: []event{{
 			UserID:    mt.userID,
 			SessionID: mt.botStartTime.Unix() * 1000, // convert to millis based on docs
 			DeviceID:  mt.deviceID,
 			EventType: eventType,
-			Props:     eventProps,
-			Version:   mt.props.CliVersion,
+			Props:     mergedProps,
+			Version:   mt.cliVersion,
```
nikhilsaraf, line 336: cliVersion is not added on the gui factory method so this will produce an empty string

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
 		})
 	}
 }
+
+func TestToMapStringInterface_SuccessMap(t *testing.T) {
+	success := map[string]interface{}{
+		"test": true,
+	}
+	m, e := ToMapStringInterface(success)
+	assert.Equal(t, success, m)
```
nikhilsaraf, line 45: we want to check for the error first and if that check fails then return
then we want to check if `m` is the same as `success`

this applies to all tests. generally if the error assertion fails then the other value(s) are invalid anyway so we don't want false negatives or false positives.

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
...
+		EmptyInts:      []int{},
+		NonEmptyInts:   []int{1, 2},
+	}
+
+	want := map[string]interface{}{
+		"EmptyInts":      []interface{}{},
+		"EmptyString":    "",
+		"EmptyStruct":    map[string]interface{}{},
+		"False":          false,
+		"Float":          1.,
```
nikhilsaraf, line 85: I know go allows this, but can we avoid it and use 1.0 instead for all places where we do this

there are a few such things that are syntactic sugar, but which make the codebase more complex for newcomers and non-go developers, so best to avoid where possible

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
...
+		Int:            1,
+		Float:          1.0,
+		EmptyStruct:    struct{}{},
+		NonEmptyStruct: NonEmptyStruct{1},
+		EmptyInts:      []int{},
+		NonEmptyInts:   []int{1, 2},
+	}
+
+	want := map[string]interface{}{
+		"EmptyInts":      []interface{}{},
```
nikhilsaraf, line 81: the order of this `want` map should be the same as the `success` struct to make it more readable and easier to follow.

nikhilsaraf: PS: since it's not intuitive without the underlying context, please add a comment explaining why we "lose the type information" in places such as `EmptyInts` and `NonEmptyStruct` as examples

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
...
+		"EmptyString":    "",
+		"EmptyStruct":    map[string]interface{}{},
+		"False":          false,
+		"Float":          1.,
+		"Int":            1.,
```
nikhilsaraf, line 86: @debnil why is the `Int` field showing up as a `float`?
is this an error somewhere and we cannot convert ints using this approach?

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
...
+		"NonEmptyInts":   []interface{}{1., 2.},
+		"NonEmptyString": "hello world",
+		"NonEmptyStruct": map[string]interface{}{"value": 1.},
+		"True":           true,
+	}
+
+	m, e := ToMapStringInterface(success)
+	assert.Equal(t, want, m)
```
nikhilsaraf, line 94: see comment above about checking error first

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
...
+	assert.NoError(t, e)
+}
+
+func TestToMapStringInterface_FailureFloat(t *testing.T) {
+	failure := 0.0
+	m, e := ToMapStringInterface(failure)
+	assert.EqualError(t, e, "could not unmarshal json to interface: json: cannot unmarshal number into Go value of type map[string]interface {}")
```
nikhilsaraf, line 101: we want the expected error to be the string you typed, and the actual error to be `e`, but the order is reversed and will produce an incorrect message on test failure

support/utils/functions_test.go
```diff
@@ -36,3 +36,90 @@ func TestDedupe(t *testing.T) {
...
+	}
+	wantMap := map[string]interface{}{
+		"1": 1,
+		"2": 2,
+		"3": 3,
+		"4": "four",
+	}
+
+	gotMap, e := MergeMaps(original, override)
+	assert.Equal(t, wantMap, gotMap)
```
nikhilsaraf, line 123: 1. check error first and return if fails (as mentioned above)
2. check that the `original` and `override` is unmodified

nikhilsaraf (dismissed):
a few concerns inline. if you want to submit as a new PR that's fine with me

Note that without the `cli_version` in the gui metrics factory method we won't have version info for the GUI.

support/utils/functions_test.go
debnil: This isn't an error, it's how JSON works. The standard only has numbers, and the `json` package assumes float64 when you haven't defined anything else (just Unmarshaling to an `interface{}` rather than a custom `struct`) - which is what we are doing. So, this is expected behavior. I've added a comment documenting this.

I think this is fine for our purposes. It would be bad if we were going from float to int (i.e., 64-bit to 32-bit precision). But, since we won't lose precision through this approach, I'm fine with it. It's good that we have a test that documents this behavior.

support/utils/functions_test.go
debnil: The code is correct - the second parameter is the error, and the third is the error string. See the ([documentation](https://godoc.org/github.com/stretchr/testify/assert#EqualError)).

debnil:
@nikhilsaraf Thanks! All of the above is addressed. Feel free to merge if it looks good to you.

support/utils/functions_test.go
nikhilsaraf: yes, that's fair and best to document this in a comment. works for me 👍

nikhilsaraf (approved):
thanks for getting this to the finish line 🎉 
now that we have this, we can start sending events by making use of `sendMetricEvent.js`.

let's have the next PR be minimal so we can set the tone for what comprises an event submission from the frontend in terms of usage of the capabilities added here. We can then replicate slowly for the various type of frontend events one by one.

Issue #483, opened 2020-08-21
[ad-hoc] [3](5) Add test for volumeFilter to ensure we correctly handle offers that will put us over


nikhilsaraf:
use tests in `dailyVolumeByDate_test.go` and `TestMakeVolumeFilterConfig` as inspiration.

PR #548, opened 2020-10-19
Implement buy twap strategy (closes #522)

debnil:
This PR implements the buy twap strategy with all necessary files and closes #522.

The following log runs confirm that the strategy successfully creates buy offers that fill previously made sell offers, through sell twap, with the same config:

```
2020/10/19 15:46:45 Starting fill tracker with 2 handlers
2020/10/19 15:46:45 Starting the trader bot...
2020/10/19 15:46:45 ----------------------------------------------------------------------------------------------------
2020/10/19 15:46:45 there were no trades, leaving lastCursor value as 5621167297667073-0
2020/10/19 15:46:46 synchronized state loading is disabled
2020/10/19 15:46:46  (base) assetA=native, maxA=8999.99997970, trustA=math.MaxFloat64
2020/10/19 15:46:46 (quote) assetB=COUPON:GBMMZMK2DC4FFP4CAI6KCVNCQ7WLO5A7DQU7EC7WGHRDQBZB763X4OQI, maxB=85.33766260, trustB=922337203685.47753906
2020/10/19 15:46:46 orderConstraints for trading pair XLM/COUPON: OrderConstraints[PricePrecision: 7, VolumePrecision: 7, MinBaseVolume: 0.0000001, MinQuoteVolume: <nil>]
2020/10/19 15:46:46 liabilities after resetting
2020/10/19 15:46:46 asset=base  , balance=8999.99997970, trust=math.MaxFloat64, minAccountBal=22.00000000, buyingLiabilities=0.00000000, sellingLiabilities=0.00000000
2020/10/19 15:46:46 asset=quote , balance=85.33766260, trust=922337203685.47753906, minAccountBal=0.08533766, buyingLiabilities=0.00000000, sellingLiabilities=0.00000000
2020/10/19 15:46:46 GetLevels, unix timestamp for 'now' in UTC = 1603147606 (2020-10-19 22:46:46.8704 +0000 UTC)
2020/10/19 15:46:46 volumeFilter = volume/daily/sell/base/10000.0/exact
2020/10/19 15:46:46 returning kraken API key at index 0
2020/10/19 15:46:47 (modifier: mid) price from exchange feed (kraken/XXLM/ZUSD/mid): bidPrice=0.085293, askPrice=0.085349, midPrice=0.085321, lastTradePrice=0.085287; price=0.085321
[...]

Issue #522, opened 2020-10-08
[8] buyTwap strategy


nikhilsaraf:
We have a sellTwap strategy, which uses the compose strategy to set the buy side as empty (delete) and sell side as the sell twap level provider.

We want an equivalent strategy that works on the buy side where buy side is set to the sell twap level provider and sell side is empty (delete). In such situations you will need to adjust the ordering of the base asset and quote asset as is done in the buysell strategy.

This should reuse the sample sellTwap config file. If there are any differences when creating the buy version please highlight in comments. If they are exactly the same we can rename the sample config file to `sample_twap.cfg` indicating that it is common for both buy and sell twap strategies.

Here are two useful resources:

1. [sellTwap strategy](https://github.com/stellar/kelp/blob/master/plugins/sellTwapStrategy.go) file
1. [buysell strategy](https://github.com/stellar/kelp/blob/master/plugins/buysellStrategy.go) file
1. [sample sellTwap config](https://github.com/stellar/kelp/blob/master/examples/configs/trader/sample_selltwap.cfg) file
1. [sample PR on how to create a new trading strategy](https://github.com/stellar/kelp/pull/494).

We want to include this in the v1.10.0 release and is a priority over the Kelp GUI rc2.

Testing is a manual process in this case, which is what contributes to 3 of the 5 points allotted to this task

## Testing

This should be tested manually on SDEX with a 10-second update interval (trader config) and a 60-second bucket interval (strategy config file).

Please upload the log file here from the test run so I can look at it and ensure the strategy is working correctly (and also for posterity).

### Triggering Trades
We want some way of triggering trades against the buyTwap strategy. One way is to create the necessary accounts on SDEX testnet and start consuming the maker orders created by the buytwap manually. You can use the testnet versions of interstellar.exchange or stellarterm for this.

PR #552, opened 2020-10-22
Add tests for the volume filter (part of #483)

debnil:
This PR adds tests for the volume filter. This lets us add tests for the volume filter, which in turn lets us make changes needed for `buyTwap` (part of #522 and #548).

This PR does not close issue #483 since we've split up the tests into two parts.


plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+		},
+	}
+
+	for _, k := range testCases {
+		t.Run(k.name, func(t *testing.T) {
+			tbb := k.dailyTBB
+			actual, e := k.filter.volumeFilterFn(k.dailyOTB, tbb, k.op)
+			assert.Equal(t, k.wantError, e)
+			assert.Equal(t, k.wantOp, actual)
+			assert.Equal(t, k.wantTBB, tbb)
```
nikhilsaraf, line 432: it's confusing that tbb is modified in volumeFilterFn -- that's my fault.
Can you please rename it in the function and here to `dailyTBBAccumulator` -- the accumulator denotes that it's a modified variable, which helps distinguish it from OTB which is not modified

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+		})
+	}
+}
+
+func createFloatPtr(x float64) *float64 {
```
nikhilsaraf, line 437: use `pointy.Float64(value)` instead -- see usage examples of the `pointy` library in filterFactory_test.go

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			name:             "failure - display base",
+			configValue:      "",
+			exchangeName:     "",
+			tradingPair:      &model.TradingPair{Base: "FAIL", Quote: ""},
+			assetDisplayFn:   testAssetDisplayFn,
+			baseAsset:        hProtocol.Asset{},
+			quoteAsset:       hProtocol.Asset{},
+			db:               db,
+			config:           &VolumeFilterConfig{},
+			wantError:        fmt.Errorf("could not convert base asset (FAIL) from trading pair via the passed in assetDisplayFn: cannot recognize the asset FAIL"),
```
nikhilsaraf, line 107: a case like this is more appropriate in the tests for asset display function, better to remove here

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			name:             "failure - query",
+			configValue:      "",
+			exchangeName:     "",
+			tradingPair:      &model.TradingPair{Base: "XLM", Quote: "XLM"},
+			assetDisplayFn:   testAssetDisplayFn,
+			baseAsset:        testNativeAsset,
+			quoteAsset:       testNativeAsset,
+			config:           &VolumeFilterConfig{},
+			db:               nil,
+			wantError:        fmt.Errorf("could not make daily volume by date Query: could not make daily volume by date action: the provided db should be non-nil"),
```
nikhilsaraf, line 132: this is a good test to have in general but since it's testing logic in the factory method for `queries.MakeDailyVolumeByDateForMarketIdsAction` it's best to leave out. It does test that we are propagating errors from subcomponents correctly, but that would result in too many tests and is better caught in other ways.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+		},
+		{
+			name:           "success - empty config",
+			configValue:    "",
+			exchangeName:   "",
+			tradingPair:    &model.TradingPair{Base: "XLM", Quote: "XLM"},
+			assetDisplayFn: testAssetDisplayFn,
+			baseAsset:      testNativeAsset,
+			quoteAsset:     testNativeAsset,
+			config:         testEmptyVolumeConfig,
```
nikhilsaraf, line 143: hmm.. I think this test should have failed. maybe you've found a bug.

nikhilsaraf: the logic in the factory method should run config.validate() to ensure this test fails.
we will need to create a config.validate() method -- which does the following (and is also tested):
1. tests that exactly one of `SellBaseAssetCapInBaseUnits` or `SellBaseAssetCapInQuoteUnits` is nil.
2. tests that `mode` is one of the two possible values it uses
3. additionalMarketIDs is non-nil
4. optionalAccountIDs is non-nil

Adding this could create conflicts with PR # 2 so can you make a note to add this as part of PR # 2 and then add this test case to ensure it fails on an invalid config?

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+		},
+		{
+			name:           "success - market ids",
+			configValue:    "",
+			exchangeName:   "",
+			tradingPair:    &model.TradingPair{Base: "XLM", Quote: "XLM"},
+			assetDisplayFn: testAssetDisplayFn,
+			baseAsset:      testNativeAsset,
+			quoteAsset:     testNativeAsset,
+			config:         &VolumeFilterConfig{additionalMarketIDs: marketIDs},
```
nikhilsaraf, line 162: most of the configs used in these test cases are invalid -- see comment above about valid configs.

plugins/volumeFilter_test.go
nikhilsaraf: see related test for this -- `TestMakeVolumeFilterConfig`. Maybe it needs a negative version of the test which fails on invalid entries

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+		dailyTBB  *VolumeFilterConfig
+		op        *txnbuild.ManageSellOffer
+		wantOp    *txnbuild.ManageSellOffer
+		wantError error
+		wantTBB   *VolumeFilterConfig
+	}{
+		{
+			name:      "failure - is selling",
+			filter:    emptyFilter,
+			dailyOTB:  nil,
```
nikhilsaraf, line 262: a test case like this should be directly on `utils.IsSelling` (see reasons mentioned above).
can you add it there? (or put in a GH issue to be added later?)

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			op:        &txnbuild.ManageSellOffer{Buying: txnbuild.NativeAsset{}, Selling: txnbuild.CreditAsset{}},
+			wantOp:    nil,
+			wantError: fmt.Errorf("error when running the isSelling check for offer '{Selling:{Code: Issuer:} Buying:{} Amount: Price: OfferID:0 SourceAccount:<nil>}': invalid assets, there are more than 2 distinct assets: sdexBase={native  }, sdexQuote={native  }, selling={ }, buying={}"),
+			wantTBB:   nil,
+		},
+		{
+			name:      "failure - invalid sell price in op",
+			filter:    emptyFilter,
+			dailyOTB:  nil,
+			dailyTBB:  nil,
```
nikhilsaraf, line 273: you've got some very comprehensive tests here! 👍 

However, that brings up challenges of it's own. I think we can take more of a whitebox testing approach where we only test the logic that is more likely to have errors, which will keep the tests a little smaller and more targeted, making them easier to reason about.

for example, it's safe to assume that the offer passed in is fully valid and has a valid price and that parse float will not fail, and that we call parsefloat on it correctly.

If we want to test the functionality more thoroughly, the best approach would be to extract the function that does the validation and extraction of values from the various inputs and write a functional test for it. this will ensure we have smaller and more targeted tests. I'm not opposed to doing this, I just don't think it's necessary yet because we have larger gaping holes in our test infrastructure that can benefit from test coverage.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			)
+
+			assert.Equal(t, k.wantError, e)
+			assert.Equal(t, k.wantSubmitFilter, actual)
+		})
+	}
+}
+
+func TestVolumeFilterFn(t *testing.T) {
+	db := connectTestDb()
```
nikhilsaraf, line 233: we don't need a db here -- we can manually construct a `volumeFilter` by passing in `nil` for the `dailyVolumeByDateQuery` since that's not used. Really what's actually used in the function is the `config` and the base/quote assets. I'm ok with changing the signature of `volumeFilterFn` to take in only those fields and making it a function that is not hanging off the volumeFilter. This will ensure we don't need to construct excessive objects to test it

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			wantOp:    nil,
+			wantError: fmt.Errorf("could not convert amount () to float: strconv.ParseFloat: parsing \"\": invalid syntax"),
+			wantTBB:   nil,
+		},
+		{
+			name:      "failure - invalid amount in op",
+			filter:    emptyFilter,
+			dailyOTB:  nil,
+			dailyTBB:  nil,
+			op:        &txnbuild.ManageSellOffer{Buying: txnbuild.NativeAsset{}, Selling: txnbuild.NativeAsset{}, Price: "0.0"},
```
nikhilsaraf, line 294: all of these tests that are testing the validity of inputs can be removed

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			wantOp:    nil,
+			wantError: fmt.Errorf("could not convert amount () to float: strconv.ParseFloat: parsing \"\": invalid syntax"),
+			wantTBB:   nil,
+		},
+		{
+			name:      "success - selling, no filter sell caps",
+			filter:    emptyFilter,
+			dailyOTB:  &VolumeFilterConfig{},
+			dailyTBB:  &VolumeFilterConfig{SellBaseAssetCapInBaseUnits: createFloatPtr(0.0), SellBaseAssetCapInQuoteUnits: createFloatPtr(0.0)},
+			op:        &txnbuild.ManageSellOffer{Buying: txnbuild.NativeAsset{}, Selling: txnbuild.NativeAsset{}, Price: "1.0", Amount: "100.0"},
```
nikhilsaraf, line 304: we should have a helper function called `makeManageSellOffer(price, volume)` so we don't have to create structs everywhere -- makes it a lot more readable

nikhilsaraf: better yet -- this should be a parameter to the testCase struct input and the test should construct the appropriate manage sell offer from these two fields of price and volume

nikhilsaraf: similarly, the inputs for otb and tbb should be:
- otbBaseCap
- otbQuoteCap
- tbbBaseCap
- tbbQuoteCap

this will make these test case structs a lot smaller and will contain less boilerplate

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+			},
+			dailyOTB: &VolumeFilterConfig{
+				SellBaseAssetCapInBaseUnits:  createFloatPtr(0.0),
+				SellBaseAssetCapInQuoteUnits: createFloatPtr(0.0),
+			},
+			dailyTBB: &VolumeFilterConfig{
+				SellBaseAssetCapInBaseUnits:  createFloatPtr(0.0),
+				SellBaseAssetCapInQuoteUnits: createFloatPtr(0.0),
+			},
+			op:        &txnbuild.ManageSellOffer{Buying: txnbuild.NativeAsset{}, Selling: txnbuild.NativeAsset{}, Price: "1.0", Amount: "100.0"},
```
nikhilsaraf, line 327: we have a price of 1.0 in a few places -- that's not very helpful since we won't do any price conversions then. It should be anything but 1.0  :)

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+				name:                   "volumeFilter",
+				configValue:            "",
+				baseAsset:              testNativeAsset,
+				quoteAsset:             testNativeAsset,
+				config:                 &VolumeFilterConfig{SellBaseAssetCapInBaseUnits: createFloatPtr(1.0), mode: volumeFilterModeExact},
+				dailyVolumeByDateQuery: dailyVolumeByDateQuery,
+			},
+			dailyOTB: &VolumeFilterConfig{
+				SellBaseAssetCapInBaseUnits:  createFloatPtr(0.0),
+				SellBaseAssetCapInQuoteUnits: createFloatPtr(0.0),
```
nikhilsaraf, line 344: I haven't checked the values on these tests yet but once we have a simpler interface to these test cases it will be a lot easier to go through. we should end up without any struct constructions in tests cases wherever possible.

nikhilsaraf (commented):
I haven't checked the values on `TestVolumeFilterFn` yet since't it's a little hard to go through.

I've made some comments on how we can have a simpler interface to these test cases, which should make it a lot easier to go through.

I've added quite a few comments, let me know if you want to discuss tomorrow to help unblock.

plugins/volumeFilter.go
```diff
@@ -90,6 +89,24 @@ func makeFilterVolume(
 	}, nil
 }
 
+func makeDailyVolumeByDateQuery(
```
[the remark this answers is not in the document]
debnil: The extraction was to generate needed queries for test cases. This avoids repeating the same 5-6 lines, when we vary the market IDs or account IDs. I didn't add the action to sell or buy in this PR because there's no notion of "buy" in the codebase yet - it would be added in the subsequent PR that tests buy out.

With that said, I don't think this matters if we constrain the test cases, as suggested below, and reduce database calls.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+	if e != nil {
+		panic(e)
+	}
+	return db
+}
+
+var testNativeAsset hProtocol.Asset = hProtocol.Asset{Type: "native"}
+
+func TestMakeFilterVolume(t *testing.T) {
+	testAssetDisplayFn := model.AssetDisplayFn(func(asset model.Asset) (string, error) {
```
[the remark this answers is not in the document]
debnil: Good catch - calling the second as a factory method to generate this test function.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,439 @@
...
+	db := connectTestDb()
+
+	emptyConfigVolumeQuery, e := makeDailyVolumeByDateQuery(db, "", "native", "native", []string{}, []string{})
+	if e != nil {
+		t.Log("could not make empty config volume query")
+		return
+	}
+
+	marketIDs := []string{"marketID"}
+	marketConfigVolumeQuery, e := makeDailyVolumeByDateQuery(db, "", "native", "native", []string{}, marketIDs)
```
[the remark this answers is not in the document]
debnil: Good call on the empty db. I'm not 100% certain that we can reconstruct the `sqlQuery` on `DailyVolumeByDate` without calling an extracted function like `makeDailyVolumeByDateQuery`, since the ultimate result looks like the below. Will come back to this.

`SELECT SUM(base_volume) as total_base_volume, SUM(counter_cost) as total_counter_volume FROM trades WHERE market_id IN ('4ba0ebdac9', 'marketID') AND account_id IN ('accountID') AND DATE(date_utc) = $1 and action = $2 group by DATE(date_utc)`

plugins/volumeFilter_test.go
debnil: That sounds good. Out of curiosity, though, how else would we catch propagating errors from subcomponents?

plugins/volumeFilter_test.go
```diff
@@ -2,224 +2,293 @@ package plugins
...
+			name: "non nil cap in base, nil cap in quote",
+			config: &VolumeFilterConfig{
+				SellBaseAssetCapInBaseUnits: pointy.Float64(1.0),
+				mode:                        volumeFilterModeExact,
+				additionalMarketIDs:         []string{},
+				optionalAccountIDs:          []string{},
+			},
+			wantError: nil,
+			wantSubmitFilter: &volumeFilter{
+				name:       "volumeFilter",
```
nikhilsaraf, line 66: name and base/quote assets are not changing across the test cases -- can we move these to be inside the test?

we also don't need to re-specify the config inside the `wantSubmitFilter` since that's exactly the same as what we pass in -- in the test we can check that `assert.Equal(t, k.config, submitFilter.config)`
in fact I think the test case should only take in 1 want value: `wantDailyVolumeByDateQuery`.
we don't need wantError either because none of our cases will produce an error

Take a look at `TestFilterOrdersByVolume` for an example of a test where the testCases does not match the function inputs and outputs exactly to make them more readable and reduce boilerplate ([link](https://github.com/stellar/kelp/blob/master/plugins/mirrorStrategy_test.go#L107))

nikhilsaraf: additionally -- we can have two test runs for each case specified here inside the `t.Run()` function below -- one with nil base cap, and the other with a nil quote cap. that way we don't need to specify 2x cases here.

we can do the same for exact and ignore -- since those aren't modified either. so this will be a nested for loop for each case specified here.

something like this:
```
for test cases:
    for _, m := range []volumeFilterMode{exact, ignore} {
        baseCapConfig := &VolumeFilterConfig{
            SellBaseAssetCapInBaseUnits: point.Float64(1.0),
            SellBaseAssetCapInQuoteUnits: nil,
            mode: m,
            additionalMarketIDs: k.additionalMarketIDs,
            optionalAccountIDs: k.optionalAccountIDs,
        }

        quoteCapConfig := &VolumeFilterConfig{
            SellBaseAssetCapInBaseUnits: nil,
            SellBaseAssetCapInQuoteUnits: point.Float64(1.0),
            mode: m,
            additionalMarketIDs: k.additionalMarketIDs,
            optionalAccountIDs: k.optionalAccountIDs,
        }

        for _, c := range []*VolumeFilterConfig{baseCapConfig, quoteCapConfig} {
            isBaseCap := c.SellBaseAssetCapInBaseUnits != nil
            name := fmt.Sprintf("%s_%s_%v", k.namePrefix, m.String(), isBaseCap)
            t.Run(name, func(t *testing.T) {
                // refer to TestFilterOrdersByVolume for how to convert test case inputs to function inputs and function outputs to test case outputs
            })
        }
    }
```

does this make sense?

nikhilsaraf: this will reduce the test cases significantly

for the empty config, we can have a special test since it doesn't conform to this model

plugins/volumeFilter_test.go
```diff
@@ -229,13 +298,31 @@ func TestMakeFilterVolume(t *testing.T) {
...
+		SellBaseAssetCapInBaseUnits:  pointy.Float64(baseCap),
+		SellBaseAssetCapInQuoteUnits: pointy.Float64(quoteCap),
+		mode:                         volumeFilterModeExact,
+		additionalMarketIDs:          []string{},
+		optionalAccountIDs:           []string{},
 	}
+}
+
+func TestVolumeFilterFn(t *testing.T) {
+	dailyVolumeByDateQuery := mustMakeTestDailyVolumeQuery([]string{}, []string{})
```
nikhilsaraf, line 325: let's hold off on changes to TestVolumeFilterFn, may be better to move it to a separate PR since it's hard to follow and many of the comments that apply to the structure of the test above apply here. I think that what we learn from an improved structure to the first test will make writing and reviewing this test much easier -- this test is also a lot more complex because it's the main functionality and deals with numbers.

nikhilsaraf (commented):
may be good to discuss over a call today, lmk

plugins/volumeFilter_test.go
```diff
@@ -2,224 +2,293 @@ package plugins
...
-		User:      os.Getenv("POSTGRES_USER"),
-		SSLEnable: false,
-	}
+var testNativeAsset hProtocol.Asset = hProtocol.Asset{Type: "native"}
 
-	_, e := postgresdb.CreateDatabaseIfNotExists(postgresDbConfig)
-	if e != nil {
-		panic(e)
-	}
+func mustMakeTestDailyVolumeQuery(optionalAccountIDs, additionalMarketIDs []string) *queries.DailyVolumeByDate {
```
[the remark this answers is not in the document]
debnil: Cool, this mostly makes sense. I think I'm just going to eliminate this function - some other refactors around helpers mean that this logic can be explicitly pushed up into `makeTestVolumeFilter`. Those changes are pushed, let me know how it looks.

plugins/volumeFilter_test.go
```diff
@@ -2,224 +2,293 @@ package plugins
...
-		baseAsset      hProtocol.Asset
-		quoteAsset     hProtocol.Asset
-		db             *sql.DB
 		config         *VolumeFilterConfig
 
 		wantError        error
 		wantSubmitFilter SubmitFilter
 	}{
-		// Note: config can be additional market IDs, make sure to test additionalMarketIds and optionalAccountIDs
+		// TODO DS Confirm the empty config fails once validation is added to the constructor
```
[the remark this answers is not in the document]
debnil: Done: https://github.com/stellar/kelp/issues/571

plugins/volumeFilter_test.go
debnil: I made some further changes that mostly simplified the interface but also rewrote `VolumeFilterFn` as a standalone function rather than a method. If you still want to hold off on changes, can do so, but might be good to just get it done in one swoop since it's massively simplified.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+}
+
+func TestMakeFilterVolume(t *testing.T) {
+	testAssetDisplayFn := model.MakeSdexMappedAssetDisplayFn(map[model.Asset]hProtocol.Asset{model.Asset("XLM"): utils.NativeAsset})
+
+	testCases := []struct {
+		name       string
+		marketIDs  []string
+		accountIDs []string
```
nikhilsaraf, line 60: test cases should take a `want` variable. the inputs should be structured so the want values are different

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,256 @@
...
+	return &VolumeFilterConfig{
+		SellBaseAssetCapInBaseUnits:  baseCapInBasePtr,
+		SellBaseAssetCapInQuoteUnits: baseCapInQuotePtr,
+		mode:                         mode,
+		additionalMarketIDs:          additionalMarketIDs,
+		optionalAccountIDs:           optionalAccountIDs,
+	}
+}
+
+func makeTestVolumeFilter(config *VolumeFilterConfig, additionalMarketIDs, optionalAccountIDs []string) *volumeFilter {
```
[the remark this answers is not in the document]
debnil: This refactor makes enough sense, I've pulled out the "someMarketID" string from this function and am defining it in the test.

plugins/volumeFilter_test.go
debnil: I don't think this is possible. the `want` variable requires an input configuration. Our current approach, which loops over modes and configs, forecloses referencing it in the `testCases` array - the config doesn't exist at this point.

Within our current test structure, we'd have to define the testCases within the inner loop of our array. We could do this, but it's a bit messy. Let me know.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,268 @@
...
+		},
+		{
+			name:       "2 market ids",
+			marketIDs:  []string{"marketID1", "marketID2"},
+			accountIDs: []string{},
+		},
+		{
+			name:       "2 dupe market ids, 1 distinct",
+			marketIDs:  []string{"marketID1", "marketID1", "marketID2"},
+			accountIDs: []string{},
```
nikhilsaraf, line 83: wantMarketIDs should be `[]string{"marketID1", "marketID2"},`

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,268 @@
...
+		},
+		{
+			name:       "1 account id",
+			marketIDs:  []string{},
+			accountIDs: []string{"accountID"},
```
nikhilsaraf, line 88: wantMarketIDs should be a non-empty string array here I think (would be the string inlined value in `firstMarketID`)?

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,268 @@
...
+				// configType is used to represent the type of config when printing test name
+				var configType string
+				if config.SellBaseAssetCapInBaseUnits != nil {
+					configType = "base"
+				} else {
+					configType = "quote"
+				}
+
+				// TODO DS Vary filter action between buy and sell, once buy logic is implemented.
+				wantFilter := makeWantVolumeFilter(config, firstMarketID, k.marketIDs, k.accountIDs, "sell")
```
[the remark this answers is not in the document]
nikhilsaraf: I get the feeling we had discussed this in the last review -- this cannot use the same `marketIDs` and `accountIDs` variable from the input to construct the `want` value.

simple example:
input marketIDs = []string{"marketID1", "marketID1"}
wantMarketIDs = []string{"marketID1"}

therefore these variables cannot be the same.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return &txnbuild.ManageSellOffer{
+		Buying:  txnbuild.NativeAsset{},
+		Selling: txnbuild.NativeAsset{},
+		Price:   price,
+		Amount:  amount,
+	}
+}
+
+func TestVolumeFilterFn(t *testing.T) {
```
nikhilsaraf, line 156: reviewing later once we have the factory method well tested -- I think we should split these up into separate PRs

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/stellar/kelp/support/utils"
+
+	"github.com/stellar/go/txnbuild"
+
+	hProtocol "github.com/stellar/go/protocols/horizon"
+	"github.com/stellar/kelp/model"
+	"github.com/stretchr/testify/assert"
+)
+
+func makeTestVolumeFilterConfig(baseCapInBase, baseCapInQuote float64, additionalMarketIDs, optionalAccountIDs []string, mode volumeFilterMode) *VolumeFilterConfig {
```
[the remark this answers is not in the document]
debnil: Yes - sorry! Some of the outstanding PRs were before we had discussed that convention.

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,268 @@
...
+	var baseCapInBasePtr *float64
+	if baseCapInBase >= 0 {
+		baseCapInBasePtr = pointy.Float64(baseCapInBase)
+	}
+
+	var baseCapInQuotePtr *float64
```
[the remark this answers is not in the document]
debnil: I don't think we decided either way. Definitely agreed that it's not amazing, but we don't have a great alternative yet, and I'm totally open to another way we can vary this.

plugins/volumeFilter_test.go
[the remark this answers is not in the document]
debnil: This doesn't do that, because the `firstMarketID` is pre-pended to `k.marketIDs` for the comparison.

The reason `wantFilter` cannot go on the test cases is here: https://github.com/stellar/kelp/pull/552#discussion_r514447104

There's no way for us to construct the `config`, which we loop over, if the `marketIDs` and `accountIDs` aren't already defined. But those are defined within the test case. And we can't construct the `want` value without the `config`. Thoughts on how to solve this?

plugins/volumeFilter_test.go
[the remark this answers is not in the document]
nikhilsaraf: I think it's ok to use pointers (via pointy) directly instead of the `-1` sentinels. I think this will help readability.

plugins/volumeFilter_test.go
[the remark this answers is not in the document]
nikhilsaraf: noting here, let's discuss on a call early next week on Monday or Tuesday and merge this PR in a pair programming session

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,161 @@
...
+					}
+
+					assert.Equal(t, wantFilter, actual)
+				})
+			}
+		}
+	}
+}
+
+func makeManageSellOffer(price, amount string) *txnbuild.ManageSellOffer {
```
nikhilsaraf, line 150: is this function unused? if so let's remove it

plugins/volumeFilter_test.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+func makeWantVolumeFilter(config *VolumeFilterConfig, marketIDs []string, accountIDs []string, action string) *volumeFilter {
+	query, e := queries.MakeDailyVolumeByDateForMarketIdsAction(&sql.DB{}, marketIDs, action, accountIDs)
+	if e != nil {
+		panic(e)
+	}
+
+	return &volumeFilter{
+		name:                   "volumeFilter",
```
nikhilsaraf, line 26: add `configValue: ""`

nikhilsaraf (commented):
should be good to go once we remove the unused function commented inline

nikhilsaraf (approved):
lgtm, thanks! 🚢

nikhilsaraf:
Note: This PR does not close issue #483 since we've split up the tests into two parts (fyi @debnil)
we should close the gh issue when we add the test for the function as well

Issue #551, opened 2020-10-22
[3] perf-based metrics


nikhilsaraf:
See "Performance Based Metrics"
https://docs.google.com/document/d/1R5hU7dSloQHZ0XNZkUrS0LPOr2gLVpteXq5NvR-Fy2k/edit#

PR #555, opened 2020-10-23
Add missing CLI metrics from inputs (part of #551)

debnil:
This PR adds some missing metrics to the CLI data tracking, using already-inputted values from the command line. This is part of #551. 

Subsequent PRs will add counters for the various metric events; hardware-based characteristics; and, later, a network event.

### DIFF
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -538,8 +538,8 @@ func runTradeCmd(options inputs) {
 		env,
 		runtime.GOOS,
 		runtime.GOARCH,
-		runtime.Version(),
 		"unknown_todo", // TODO DS Determine how to get GOARM.
+		runtime.Version(),
 		guiVersionFlag,
 		*options.strategy,
 		botConfig.TickIntervalSeconds,
@@ -559,10 +559,10 @@ func runTradeCmd(options inputs) {
 		int(botConfig.MonitoringPort) != 0,
 		len(botConfig.Filters) > 0,
 		botConfig.PostgresDbConfig != nil,
+		*options.logPrefix != "",
 		*options.operationalBuffer,
 		*options.operationalBufferNonNativePct,
 		*options.simMode,
-		*options.logPrefix,
 		*options.fixedIterations,
 	)
 	if e != nil {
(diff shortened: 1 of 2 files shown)


cmd/trade.go
```diff
@@ -538,6 +538,7 @@ func runTradeCmd(options inputs) {
 		env,
 		runtime.GOOS,
 		runtime.GOARCH,
+		runtime.Version(),
```
nikhilsaraf, line 541: I think this has been passed in for the `goarm` param instead of the `go_version` param

support/metrics/metricsTracker.go
```diff
@@ -75,6 +76,11 @@ type commonProps struct {
 	EnabledFeatureMonitoring         bool    `json:"enabled_feature_monitoring"`
 	EnabledFeatureFilters            bool    `json:"enabled_feature_filters"`
 	EnabledFeaturePostgres           bool    `json:"enabled_feature_postgres"`
+	OperationalBuffer                float64 `json:"operational_buffer"`
+	OperationalBufferNonNativePct    float64 `json:"operational_buffer_non_native_pct"`
+	SimMode                          bool    `json:"sim_mode"`
+	LogPrefix                        string  `json:"log_prefix"`
```
nikhilsaraf, line 82: `logPrefix` should be a `bool` -- we want to only know if they have enabled the feature.
in fact, can we put this above operationalBuffer and call this `EnabledFeatureLogging`?

cmd/trade.go
```diff
@@ -558,6 +559,11 @@ func runTradeCmd(options inputs) {
 		int(botConfig.MonitoringPort) != 0,
 		len(botConfig.Filters) > 0,
 		botConfig.PostgresDbConfig != nil,
+		*options.operationalBuffer,
+		*options.operationalBufferNonNativePct,
+		*options.simMode,
+		*options.logPrefix,
```
nikhilsaraf, line 565: logPrefix bool -- see comment below for details

nikhilsaraf (commented):
comments inline

nikhilsaraf (approved):
looks great, thanks! 🚢

PR #573, opened 2020-10-30
Add metrics for operation counts (part of #551)

debnil:
This PR adds metrics for operation counts. It's marked as a draft PR because I have a few outstanding questions, and haven't fully tested it until these are resolved.

1. Does this implementation and architecture generally make sense?
2. Am I passing the correct parameters for prune ops and number of update ops in update?
3. Where is the number of update ops in delete supported to be? The only other len(ops) check in `trader/trader.go` is for deletion, but from the data doc, it looks like this count is for the update event, not deletion.

### DIFF
--- a/support/metrics/metricsTracker.go
+++ b/support/metrics/metricsTracker.go
@@ -31,7 +31,6 @@ type MetricsTracker struct {
 	botStartTime        time.Time
 	isDisabled          bool
 	updateEventSentTime *time.Time
-	numOps              numOps
 }
 
 // TODO DS Investigate other fields to add to this top-level event.
@@ -109,10 +108,12 @@ type eventWrapper struct {
 	Events []event `json:"events"`
 }
 
-type numOps struct {
-	PruneOps         int `json:"num_prune_ops"`
-	UpdatesOpsDelete int `json:"num_update_ops_delete"`
-	UpdatesOpsUpdate int `json:"num_update_ops_update"`
+// UpdateLoopResult contains the results of the orderbook update.
+type UpdateLoopResult struct {
+	Success            bool
+	NumPruneOps        int
+	NumUpdateOpsDelete int
+	NumUpdateOpsUpdate int
 }
 
 // response structure taken from here: https://help.amplitude.com/hc/en-us/articles/360032842391-HTTP-API-V2#tocSsuccesssummary
@@ -215,27 +216,7 @@ func MakeMetricsTracker(
 		botStartTime:        botStartTime,
 		isDisabled:          isDisabled,
 		updateEventSentTime: nil,
-		numOps: numOps{
-			PruneOps:         0,
-			UpdatesOpsDelete: 0,
-			UpdatesOpsUpdate: 0,
-		},
 	}, nil
-}
-
-// AddPruneOps increments the total prune ops by the number of new ops.
-func (mt *MetricsTracker) AddPruneOps(numNewOps int) {
-	mt.numOps.PruneOps += numNewOps
-}
-
-// AddUpdatesOpsDelete increments the total update ops from deletion, by the number of new ops.
-func (mt *MetricsTracker) AddUpdatesOpsDelete(numNewOps int) {
-	mt.numOps.UpdatesOpsDelete += numNewOps
-}
-
-// AddUpdatesOpsUpdate increments the total update ops from updates, by the number of new ops.
-func (mt *MetricsTracker) AddUpdatesOpsUpdate(numNewOps int) {
-	mt.numOps.UpdatesOpsUpdate += numNewOps
 }
 
 // GetUpdateEventSentTime gets the last sent time of the update event.
@@ -249,7 +230,7 @@ func (mt *MetricsTracker) SendStartupEvent() error {
 }
 
 // SendUpdateEvent sends the update Amplitude event.
-func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisForUpdate int64) error {
+func (mt *MetricsTracker) SendUpdateEvent(now time.Time, updateResult UpdateLoopResult, millisForUpdate int64) error {
 	commonProps := mt.props
 	commonProps.SecondsSinceStart = now.Sub(mt.botStartTime).Seconds()
 
@@ -261,12 +242,12 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
 	}
 	updateProps := updateProps{
 		commonProps:                  commonProps,
-		Success:                      success,
+		Success:                      updateResult.Success,
 		MillisForUpdate:              millisForUpdate,
 		SecondsSinceLastUpdateMetric: secondsSinceLastUpdateMetric,
-		NumPruneOps:                  mt.numOps.PruneOps,
-		NumUpdatesOpsDelete:          mt.numOps.UpdatesOpsDelete,
-		NumUpdatesOpsUpdate:          mt.numOps.UpdatesOpsUpdate,
+		NumPruneOps:                  updateResult.NumPruneOps,
+		NumUpdatesOpsDelete:          updateResult.NumUpdateOpsDelete,
+		NumUpdatesOpsUpdate:          updateResult.NumUpdateOpsUpdate,
 	}
 	e := mt.sendEvent(updateEventName, updateProps)
 	if e != nil {
--- a/support/metrics/metricsTracker.go
+++ b/support/metrics/metricsTracker.go
@@ -31,6 +31,7 @@ type MetricsTracker struct {
 	botStartTime        time.Time
 	isDisabled          bool
 	updateEventSentTime *time.Time
+	numOps              numOps
 }
 
 // TODO DS Investigate other fields to add to this top-level event.
@@ -89,6 +90,9 @@ type updateProps struct {
 	Success                      bool    `json:"success"`
 	MillisForUpdate              int64   `json:"millis_for_update"`
 	SecondsSinceLastUpdateMetric float64 `json:"seconds_since_last_update_metric"` // helps understand total runtime of bot when summing this field across events
+	NumPruneOps                  int     `json:"num_prune_ops"`
+	NumUpdatesOpsDelete          int     `json:"num_update_ops_delete"`
+	NumUpdatesOpsUpdate          int     `json:"num_update_ops_update"`
 }
 
 // deleteProps holds the properties for the delete Amplitude event.
@@ -103,6 +107,12 @@ type deleteProps struct {
 type eventWrapper struct {
 	ApiKey string  `json:"api_key"`
 	Events []event `json:"events"`
+}
+
+type numOps struct {
+	PruneOps         int `json:"num_prune_ops"`
+	UpdatesOpsDelete int `json:"num_update_ops_delete"`
+	UpdatesOpsUpdate int `json:"num_update_ops_update"`
 }
 
 // response structure taken from here: https://help.amplitude.com/hc/en-us/articles/360032842391-HTTP-API-V2#tocSsuccesssummary
@@ -205,7 +215,27 @@ func MakeMetricsTracker(
 		botStartTime:        botStartTime,
 		isDisabled:          isDisabled,
 		updateEventSentTime: nil,
+		numOps: numOps{
+			PruneOps:         0,
+			UpdatesOpsDelete: 0,
+			UpdatesOpsUpdate: 0,
+		},
 	}, nil
+}
+
+// AddPruneOps increments the total prune ops by the number of new ops.
+func (mt *MetricsTracker) AddPruneOps(numNewOps int) {
+	mt.numOps.PruneOps += numNewOps
+}
+
+// AddUpdatesOpsDelete increments the total update ops from deletion, by the number of new ops.
+func (mt *MetricsTracker) AddUpdatesOpsDelete(numNewOps int) {
+	mt.numOps.UpdatesOpsDelete += numNewOps
+}
+
+// AddUpdatesOpsUpdate increments the total update ops from updates, by the number of new ops.
+func (mt *MetricsTracker) AddUpdatesOpsUpdate(numNewOps int) {
+	mt.numOps.UpdatesOpsUpdate += numNewOps
 }
 
 // GetUpdateEventSentTime gets the last sent time of the update event.
@@ -234,6 +264,9 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
 		Success:                      success,
 		MillisForUpdate:              millisForUpdate,
 		SecondsSinceLastUpdateMetric: secondsSinceLastUpdateMetric,
+		NumPruneOps:                  mt.numOps.PruneOps,
+		NumUpdatesOpsDelete:          mt.numOps.UpdatesOpsDelete,
+		NumUpdatesOpsUpdate:          mt.numOps.UpdatesOpsUpdate,
 	}
 	e := mt.sendEvent(updateEventName, updateProps)
 	if e != nil {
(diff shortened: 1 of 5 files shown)


trader/trader.go
```diff
@@ -401,6 +401,8 @@ func (t *Trader) update() bool {
 			return false
 		}
 
+		t.metricsTracker.AddPruneOps(len(pruneOps))
```
nikhilsaraf, line 404: we shouldn't update the `metricsTracker` with this because the metrics tracker is only responsible to send metrics, not to accumulate metrics.

the `update()` function should return a struct like this:
```
{
success: bool
numPruneOps: int
numUpdateOpsDelete: int
numUpdateOpsUpdate: int
}
```

then in the caller, we are already sending `sendUpdateEvent` -- this should take in this new struct instead of taking in the `success` field which is currently a `bool`.

does this make sense?

support/metrics/metricsTracker.go
```diff
@@ -205,9 +215,29 @@ func MakeMetricsTracker(
...
+		numOps: numOps{
+			PruneOps:         0,
+			UpdatesOpsDelete: 0,
+			UpdatesOpsUpdate: 0,
+		},
 	}, nil
 }
 
+// AddPruneOps increments the total prune ops by the number of new ops.
+func (mt *MetricsTracker) AddPruneOps(numNewOps int) {
```
nikhilsaraf, line 227: we don't need these functions

support/metrics/metricsTracker.go
```diff
@@ -205,9 +215,29 @@ func MakeMetricsTracker(
 		botStartTime:        botStartTime,
 		isDisabled:          isDisabled,
 		updateEventSentTime: nil,
+		numOps: numOps{
```
nikhilsaraf, line 218: don't need this

support/metrics/metricsTracker.go
```diff
@@ -105,6 +109,12 @@ type eventWrapper struct {
 	Events []event `json:"events"`
 }
 
+type numOps struct {
```
nikhilsaraf, line 112: don't need this

support/metrics/metricsTracker.go
```diff
@@ -89,6 +90,9 @@ type updateProps struct {
 	Success                      bool    `json:"success"`
 	MillisForUpdate              int64   `json:"millis_for_update"`
 	SecondsSinceLastUpdateMetric float64 `json:"seconds_since_last_update_metric"` // helps understand total runtime of bot when summing this field across events
+	NumPruneOps                  int     `json:"num_prune_ops"`
```
nikhilsaraf, line 93: we need this so we can leave it here

support/metrics/metricsTracker.go
```diff
@@ -31,6 +31,7 @@ type MetricsTracker struct {
 	botStartTime        time.Time
 	isDisabled          bool
 	updateEventSentTime *time.Time
+	numOps              numOps
```
nikhilsaraf, line 34: remove

support/metrics/metricsTracker.go
```diff
@@ -234,6 +264,9 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
 		Success:                      success,
 		MillisForUpdate:              millisForUpdate,
 		SecondsSinceLastUpdateMetric: secondsSinceLastUpdateMetric,
+		NumPruneOps:                  mt.numOps.PruneOps,
```
nikhilsaraf, line 267: `SendUpdateEvent` should take in the struct `UpdateLoopResult` (defined in a comment below)

so here we would say:

```
Success: updateLoopResult.success,
NumPruneOps: updateLoopResult.numPruneOps,
NumUpdateOpsDelete: updateLoopResult.numUpdateOpsDelete,
NumUpdateOpsUpdate: updateLoopResult.numUpdateOpsUpdate,
```

makes sense?

nikhilsaraf (commented):
draft PR is super helpful here too, added comments for a direction we can take this in.

debnil:
@nikhilsaraf: Super helpful review! I've addressed it. Big outstanding question: am I defining the various `Num*Ops` in the right places in the `Trader.update()` method?

support/metrics/metricsTracker.go
```diff
@@ -105,6 +108,14 @@ type eventWrapper struct {
 	Events []event `json:"events"`
 }
 
+// UpdateLoopResult contains the results of the orderbook update.
+type UpdateLoopResult struct {
```
nikhilsaraf, line 112: structure of the struct looks right.
this should be accessible by the `update()` function as it's the result of that function. should live in `trader.go`
(see review about trade metrics handler mentioning that we should move this file metricsTracker.go to the plugins package)

support/metrics/metricsTracker.go
```diff
@@ -231,9 +242,12 @@ func (mt *MetricsTracker) SendUpdateEvent(now time.Time, success bool, millisFor
 	}
 	updateProps := updateProps{
 		commonProps:                  commonProps,
-		Success:                      success,
+		Success:                      updateResult.Success,
```
nikhilsaraf, line 245: these look right 👍

trader/trader.go
```diff
@@ -123,12 +123,12 @@ func (t *Trader) Start() {
 	for {
 		currentUpdateTime := time.Now()
 		if lastUpdateTime.IsZero() || t.timeController.ShouldUpdate(lastUpdateTime, currentUpdateTime) {
-			success := t.update()
+			updateResult := t.update()
```
nikhilsaraf, line 126: since the updateResult is returned by this update() function, the struct should be specified in this file trader.go (repeating this comment from above to as a point of reference for usage of this struct here)

trader/trader.go
```diff
@@ -377,28 +387,44 @@ func (t *Trader) update() bool {
...
+			NumUpdateOpsUpdate: numUpdateOpsUpdate,
+		}
 	}
 
 	// delete excess offers
 	var pruneOps []build.TransactionMutator
 	pruneOps, t.buyingAOffers, t.sellingAOffers = t.strategy.PruneExistingOffers(t.buyingAOffers, t.sellingAOffers)
-	log.Printf("created %d operations to prune excess offers\n", len(pruneOps))
-	if len(pruneOps) > 0 {
+	numPruneOps = len(pruneOps)
```
nikhilsaraf, line 414: 👍

trader/trader.go
```diff
@@ -448,20 +490,37 @@ func (t *Trader) update() bool {
...
-			return false
+			return metrics.UpdateLoopResult{
+				Success:            false,
+				NumPruneOps:        numPruneOps,
+				NumUpdateOpsDelete: numUpdateOpsDelete,
+				NumUpdateOpsUpdate: numUpdateOpsUpdate,
+			}
 		}
+
+		// t.metricsTracker.AddUpdatesOpsUpdate(len(ops))
```
nikhilsaraf, line 501: can remove commented out code

trader/trader.go
```diff
@@ -432,12 +468,18 @@ func (t *Trader) update() bool {
...
+				NumPruneOps:        numPruneOps,
+				NumUpdateOpsDelete: numUpdateOpsDelete,
+				NumUpdateOpsUpdate: numUpdateOpsUpdate,
+			}
 		}
 	}
 
-	log.Printf("created %d operations to update existing offers\n", len(ops))
-	if len(ops) > 0 {
+	numUpdateOpsUpdate = len(ops)
```
nikhilsaraf, line 480: we haven't used `numUpdateOpsDelete` anywhere.

the numUpdateOps is broken down into two subparts: numUpdateOpsUpdate and numUpdateOpsDelete.
i.e. these update ops can contain ops to create, modify, or delete offers.
(on second thought we need to accommodate the "create" case as well so need to add that metric to the UpdateLoopResult and therefore in amplitude as well).

we need a for loop over all the update ops to count them towards either `create`, `update`, or `delete` ops.
you can call that function "countUpdateOps` as an example and have that function independently tested.

nikhilsaraf (commented):
need to make one addition to count the update ops correctly and add a relevant test, but looks good for the most part. 👍 

can remove from draft status so we can merge in next update.

support/metrics/metricsTracker.go
debnil: Even if we move `metricsTracker` to plugins, this struct cannot live in `trader.go`. That creates an import cycle, since `trader.go` already imports `plugins`. So, I think this struct has to live in `plugins/metricsTracker` (or somewhere else) to avoid cycles.

trader/trader.go
debnil: This will create an import cycle, since `trader.go` already imports `plugins`, and we pass a `UpdateLoopResult` to `SendUpdateEvent`.

nikhilsaraf: ok good point. Then let's leave as-is. Can you add a comment on the declaration of UpdateLoopResult in plugins/metricsTracker pointing this out as the reason why we define it there as opposed to in trader.go?

trader/trader.go
```diff
@@ -443,24 +493,54 @@ func (t *Trader) update() bool {
...
+			return plugins.UpdateLoopResult{
+				Success:            false,
+				NumPruneOps:        numPruneOps,
+				NumUpdateOpsDelete: numUpdateOpsDelete,
+				NumUpdateOpsUpdate: numUpdateOpsUpdate,
+				NumUpdateOpsCreate: numUpdateOpsCreate,
+			}
+		}
 
+		numUpdateOpsCreate, numUpdateOpsDelete, numUpdateOpsUpdate, e = t.countOfferChangeTypes(ops)
```
nikhilsaraf, line 508: we should run this on the `opsOld` value (i.e. after we call `t.strategy.UpdateWithOps(t.buyingAOffers, t.sellingAOffers`) instead of here.

trader/trader.go
```diff
@@ -534,3 +614,31 @@ func (t *Trader) getExistingOffers() ([]hProtocol.Offer /*sellingAOffers*/, []hP
 func (t *Trader) setExistingOffers(sellingAOffers []hProtocol.Offer, buyingAOffers []hProtocol.Offer) {
 	t.sellingAOffers, t.buyingAOffers = sellingAOffers, buyingAOffers
 }
+
+func (t *Trader) countOfferChangeTypes(offers []txnbuild.Operation) (numCreate int, numUpdate int, numDelete int, e error) {
```
nikhilsaraf, line 618: can you add a test for this?

- we can pass in 3 create offers, 2 updates, 1 delete and it should recognize it.
- then we should mix the order like (1 delete, 1 updates, 1 delete, 1 create, 2 update)
- a third case with just 2 update orders
- and a fourth case with just 1 create order

nikhilsaraf: note that once we move the function call to the correct place, the input will be of type `[]build.TransactionMutator` and not `[]txnbuild.Operation`.

trader/trader.go
```diff
@@ -534,3 +614,31 @@ func (t *Trader) getExistingOffers() ([]hProtocol.Offer /*sellingAOffers*/, []hP
...
+		if e != nil {
+			return 0, 0, 0, fmt.Errorf("could not convert op to offer: %s", e)
+		}
+
+		manageSellOp := offerXdr.Body.ManageSellOfferOp
+		if manageSellOp == nil {
+			return 0, 0, 0, fmt.Errorf("offers were not of expected type ManageSellOffer")
+		}
+
+		// 0 amount represents deletion
```
nikhilsaraf, line 631: 💯

trader/trader.go
```diff
@@ -534,3 +614,31 @@ func (t *Trader) getExistingOffers() ([]hProtocol.Offer /*sellingAOffers*/, []hP
...
+		if manageSellOp.Amount == 0 {
+			numDelete++
+		} else if manageSellOp.OfferId == 0 {
+			numCreate++
+		} else {
+			numUpdate++
+		}
+	}
+
+	return
```
nikhilsaraf, line 643: code convention: we try to avoid the implicit return syntax to keep things simple by being explicit.

can you `return numCreate, numUpdate, numDelete`?

nikhilsaraf:
@debnil can take this out of draft status

trader/trader.go
```diff
@@ -615,30 +616,29 @@ func (t *Trader) setExistingOffers(sellingAOffers []hProtocol.Offer, buyingAOffe
 	t.sellingAOffers, t.buyingAOffers = sellingAOffers, buyingAOffers
 }
 
-func (t *Trader) countOfferChangeTypes(offers []txnbuild.Operation) (numCreate int, numUpdate int, numDelete int, e error) {
-	var offerXdr xdr.Operation
+func countOfferChangeTypes(offers []*txnbuild.ManageSellOffer) (int, int, int, error) {
+	numCreate, numDelete, numUpdate := 0, 0, 0
```
nikhilsaraf, line 620: it's ok to leave the signature as `(numCreate int, numUpdate int, numDelete int, e error)` so we know what each returned value is. makes it easy for readers/consumers of the function and when figuring out what to return.

just because we are explicitly returning the values doesn't mean we cannot have named variables in the return types. 
alternatively we an use this approach for the return types `(/*numCreate*/ int, /*numUpdate*/ int, /*numDelete*/ int, error)` (error doesn't need a comment in this case)
we use both of the above patterns in a few parts of the codebase

trader/trader.go
```diff
@@ -615,30 +616,29 @@ func (t *Trader) setExistingOffers(sellingAOffers []hProtocol.Offer, buyingAOffe
...
 	for _, o := range offers {
-		offerXdr, e = o.BuildXDR()
-		if e != nil {
-			return 0, 0, 0, fmt.Errorf("could not convert op to offer: %s", e)
+		if o == nil {
+			return 0, 0, 0, fmt.Errorf("offers were not of expected type ManageSellOffer")
```
nikhilsaraf, line 623: we should provide more context on the failure so it's easier to debug. such as:
`fmt.Errorf("offer at index %d was not of expected type ManageSellOffer (actual type = %T): %+v", i, o, o)`

trader/trader.go
```diff
@@ -615,30 +616,29 @@ func (t *Trader) setExistingOffers(sellingAOffers []hProtocol.Offer, buyingAOffe
...
 		}
 
-		manageSellOp := offerXdr.Body.ManageSellOfferOp
-		if manageSellOp == nil {
-			return 0, 0, 0, fmt.Errorf("offers were not of expected type ManageSellOffer")
+		opAmount, e := strconv.ParseFloat(o.Amount, 64)
+		if e != nil {
+			return 0, 0, 0, fmt.Errorf(" invalid operation amount (%s) could not be parsed as float", o.Amount)
```
nikhilsaraf, line 628: `..for operation at index %d: %v", ..., i, o` so we print the whole offer as well for easier debugging

trader/trader.go
nikhilsaraf: note that in the previous commit this was create, update, delete, and now we return create, delete, update -- this is very easy to do because the types are same which is where the above convention helps just a little

trader/trader_test.go
```diff
@@ -247,3 +250,100 @@ func TestShouldSendUpdateMetric_NilLastMetricUpdate(t *testing.T) {
 	shouldUpdate := shouldSendUpdateMetric(now, now, nil)
 	assert.Equal(t, true, shouldUpdate)
 }
+
+func TestCountOfferChangeTypes(t *testing.T) {
```
nikhilsaraf, line 254: nailed it 👍

nikhilsaraf (commented):
minor style suggestion, perfect otherwise! 👍

trader/trader.go
debnil: Great suggestion! I changed it to create-delete-update to be in alphabetical order - that felt the least arbitrary - but definitely implementing this approach to naming.

nikhilsaraf (approved):
fyi @debnil added some small consistency related adjustments

PR #574, opened 2020-11-03
Add metrics for trade tracking.

debnil:
This PR adds metrics to track trades. Thus far, it adds a handler interface and implementation. The metrics will be actually added after #508 is merged, because the architecture around adding individual metrics changes.

### DIFF
--- a/api/exchange.go
+++ b/api/exchange.go
@@ -237,6 +237,18 @@ type ExchangeShim interface {
 	FillTrackable
 }
 
+// TradeMetricsHandler is invoked by the MetricsTracker to process new trades
+type TradeMetricsHandler interface {
+	HandleTrade(trade model.Trade) error
+	Read(trades []model.Trade)
+	Reset()
+}
+
+// MetricsTracker knows how to track metrics, including trades
+type MetricsTracker interface {
+	RegisterHandler(handler TradeMetricsHandler)
+}
+
 // ConvertOperation2TM is a temporary adapter to support transitioning from the old Go SDK to the new SDK without having to bump the major version
 func ConvertOperation2TM(ops []txnbuild.Operation) []build.TransactionMutator {
 	muts := []build.TransactionMutator{}
--- a/api/exchange.go
+++ b/api/exchange.go
@@ -239,9 +239,10 @@ type ExchangeShim interface {
 
 // TradeMetricsHandler is invoked by the MetricsTracker to process new trades
 type TradeMetricsHandler interface {
-	HandleTrade(trade model.Trade) error
 	Read(trades []model.Trade)
 	Reset()
+	TotalBaseVolume() float64
+	NumTrades() int
 }
 
 // MetricsTracker knows how to track metrics, including trades
--- a/cmd/trade.go
+++ b/cmd/trade.go
@@ -609,6 +609,9 @@ func runTradeCmd(options inputs) {
 		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker: %s", e))
 	}
 
+	tradeMetricsHandler := plugins.MakeTradeMetricsHandler()
+	metricsTracker.RegisterHandler(tradeMetricsHandler)
+
 	e = metricsTracker.SendStartupEvent()
 	if e != nil {
 		logger.Fatal(l, fmt.Errorf("could not send startup event metric: %s", e))
--- a/plugins/tradeMetricsHandler.go
+++ b/plugins/tradeMetricsHandler.go
@@ -22,7 +22,7 @@ func (h *TradeMetricsHandler) Reset() {
 }
 
 // Read stores new trades internally.
-func (h *TradeMetricsHandler) Read(newTrades []model.Trade) error {
+func (h *TradeMetricsHandler) Read(newTrades []model.Trade) {
 	for _, nt := range newTrades {
 		h.trades = append(h.trades, nt)
 	}
--- a/plugins/tradeMetricsHandler.go
+++ b/plugins/tradeMetricsHandler.go
@@ -28,14 +28,15 @@ func (h *TradeMetricsHandler) Read(newTrades []model.Trade) {
 	}
 }
 
-// Get returns the number of trades.
-func (h *TradeMetricsHandler) Get() int {
+// NumTrades returns the number of trades.
+func (h *TradeMetricsHandler) NumTrades() int {
 	return len(h.trades)
 }
 
-// HandleTrade impl
-func (h *TradeMetricsHandler) HandleTrade(trade model.Trade) error {
-	// TODO: Add more if needed.
-	h.trades = append(h.trades, trade)
-	return nil
+// TotalBaseVolume returns the total base volume.
+func (h *TradeMetricsHandler) TotalBaseVolume() (total float64) {
+	for _, t := range h.trades {
+		total += t.Volume.AsFloat()
+	}
+	return
 }
--- a/plugins/tradeMetricsHandler.go
+++ b/plugins/tradeMetricsHandler.go
@@ -1,18 +1,41 @@
-package metrics
+package plugins
 
-// TODO DS Rework handler to accumulate trades, not just count.
-type tradeMetricsHandler struct {
-	numTrades int
+import (
+	"github.com/stellar/kelp/model"
+)
+
+// TradeMetricsHandler tracks the number of trades
+type TradeMetricsHandler struct {
+	trades []model.Trade
 }
 
-func (h *tradeMetricsHandler) Reset() {
-	h.numTrades = 0
+// MakeTradeMetricsHandler is a factory method for the TradeMetricsHandler
+func MakeTradeMetricsHandler() *TradeMetricsHandler {
+	return &TradeMetricsHandler{
+		trades: []model.Trade{},
+	}
 }
 
-func (h *tradeMetricsHandler) Add(numNewTrades int) {
-	h.numTrades += numNewTrades
+// Reset sets the handler's trade counter to zero.
+func (h *TradeMetricsHandler) Reset() {
+	h.trades = []model.Trade{}
 }
 
-func (h *tradeMetricsHandler) Get() int {
-	return h.numTrades
+// Read stores new trades internally.
+func (h *TradeMetricsHandler) Read(newTrades []model.Trade) error {
+	for _, nt := range newTrades {
+		h.trades = append(h.trades, nt)
+	}
+}
+
+// Get returns the number of trades.
+func (h *TradeMetricsHandler) Get() int {
+	return len(h.trades)
+}
+
+// HandleTrade impl
+func (h *TradeMetricsHandler) HandleTrade(trade model.Trade) error {
+	// TODO: Add more if needed.
+	h.trades = append(h.trades, trade)
+	return nil
 }
--- a/plugins/tradeMetricsHandler.go
+++ b/plugins/tradeMetricsHandler.go
@@ -1,0 +1,18 @@
+package metrics
+
+// TODO DS Rework handler to accumulate trades, not just count.
+type tradeMetricsHandler struct {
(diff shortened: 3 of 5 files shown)


debnil:
@nikhilsaraf WIP PR for trade metrics. The actual metrics are not added, since we change how metrics are added in the existing PRs, but all the functions to generate those metrics already exist - they just need to be called. Let me know if this handler architecture is what you had in mind when you suggested `fillHandler` and `fillTracker` as references.

api/exchange.go
```diff
@@ -237,6 +237,19 @@ type ExchangeShim interface {
 	FillTrackable
 }
 
+// TradeMetricsHandler is invoked by the MetricsTracker to process new trades
+type TradeMetricsHandler interface {
```
nikhilsaraf, line 241: not sure I understand why this is an interface. I think this should just be a concrete struct implementing `api.FillHandler`

api/exchange.go
```diff
@@ -237,6 +237,19 @@ type ExchangeShim interface {
...
+	Read(trades []model.Trade)
+	Reset()
+	TotalBaseVolume() float64
+	NumTrades() int
+}
+
+// MetricsTracker knows how to track metrics, including trades
+type MetricsTracker interface {
```
nikhilsaraf, line 249: I don't think this should be it's own interface either. we take care of handler registrations automatically in `makeFillTracker()` by calling `fillTracker.RegisterHandler(fillLogger)` -- where fillLogger can be our single instance of `tradeMetricsHandler`.

cmd/trade.go
```diff
@@ -609,6 +609,9 @@ func runTradeCmd(options inputs) {
 		logger.Fatal(l, fmt.Errorf("could not generate metrics tracker: %s", e))
 	}
 
+	tradeMetricsHandler := plugins.MakeTradeMetricsHandler()
+	metricsTracker.RegisterHandler(tradeMetricsHandler)
```
nikhilsaraf, line 613: the registration should happen closer to where the other handlers are registered in `makeFillTracker`

plugins/tradeMetricsHandler.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stellar/kelp/model"
+)
+
+// TradeMetricsHandler tracks the number of trades
+type TradeMetricsHandler struct {
+	trades []model.Trade
+}
+
+// MakeTradeMetricsHandler is a factory method for the TradeMetricsHandler
+func MakeTradeMetricsHandler() *TradeMetricsHandler {
```
nikhilsaraf, line 13: 👍

plugins/tradeMetricsHandler.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+// Reset sets the handler's trade counter to zero.
+func (h *TradeMetricsHandler) Reset() {
+	h.trades = []model.Trade{}
+}
+
+// Read stores new trades internally.
+func (h *TradeMetricsHandler) Read(newTrades []model.Trade) {
```
nikhilsaraf, line 25: Instead of calling this function `Read` we should call it `HandleFill` so it can implement `api.FillHandler`
also it should take in a single `model.Trade` and not an array

plugins/tradeMetricsHandler.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, nt := range newTrades {
+		h.trades = append(h.trades, nt)
+	}
+}
+
+// NumTrades returns the number of trades.
+func (h *TradeMetricsHandler) NumTrades() int {
```
nikhilsaraf, line 32: 👍

plugins/tradeMetricsHandler.go
```diff
@@ -0,0 +1,42 @@
...
+	return len(h.trades)
+}
+
+// TotalBaseVolume returns the total base volume.
+func (h *TradeMetricsHandler) TotalBaseVolume() (total float64) {
```
nikhilsaraf, line 37: 👍

nikhilsaraf: we should add additional metrics here -- maybe can return a map instead of individual metrics and call the function `ComputeMetrics` or something similar to indicate that we are (a) running calculations and (b) we will return the metrics result that we wanted to collect from this handler.

(Note: base units is `model.Trade.Amount` and quote units is `model.Trade.Amount * model.Trade.Price`)
- totalBaseVolume (buy and sell action are both +ve) -- buying $100 and selling $100 should equal $200
- totalQuoteVolume (buy and sell action are both +ve)
- netBaseVolume (buy action is +ve, sell is -ve) -- buying $100 and selling $100 should equal $0
- netQuoteVolume (buy action is -ve, sell action is +ve)
- numTrades
- avgTradeSizeBase (totalBaseVolume/numTrades)
- avgTradeSizeQuote (totalQuoteVolume/numTrades)
- avgTradePrice (simple average of the trade price)
- [vwap](https://en.wikipedia.org/wiki/Volume-weighted_average_price) (volume-weighted average of the trade price) -- I think this simplifies to `totalQuoteVolume/totalBaseVolume` but you'll have to confirm

anything else you can think of that would be important to capture for trades?

support/metrics/metricsTracker.go
```diff
@@ -31,6 +33,9 @@ type MetricsTracker struct {
 	botStartTime        time.Time
 	isDisabled          bool
 	updateEventSentTime *time.Time
+
+	// uninitialized
+	handlers []api.TradeMetricsHandler
```
nikhilsaraf, line 38: this should take in a single pre-initialized pointer to an instance of `tradeMetricsHandler`

support/metrics/metricsTracker.go
```diff
@@ -81,6 +86,8 @@ type commonProps struct {
 	OperationalBufferNonNativePct    float64 `json:"operational_buffer_non_native_pct"`
 	SimMode                          bool    `json:"sim_mode"`
 	FixedIterations                  uint64  `json:"fixed_iterations"`
+	NumTradesSinceLastUpdate         int     `json:"num_trades_since_last_update"`
```
nikhilsaraf, line 89: we don't need to explicitly specify these fields for now --we can just merge the map we get from the `tradeMetricsHandler`

support/metrics/metricsTracker.go
```diff
@@ -304,3 +311,41 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
 	}
 	return nil
 }
+
+// RegisterHandler adds an internal handler.
+func (mt *MetricsTracker) RegisterHandler(handler api.TradeMetricsHandler) {
```
nikhilsaraf, line 316: we don't need any of these functions here.

instead inside sendUpdate we can call `ComputeMetrics` on the `tradeMetricsHandler` instance that we hold and merge that output with the update map that we are sending to `sendEvent` -- that's it!

plugins/tradeMetricsHandler.go
nikhilsaraf: PS: we will need to combine this with `Reset()` and make sure that we create a copy of the trades structure and then reset it, and then run our computations. This will ensure that there is a very small window in which we lose trades. We should also add a TODO NS at that point so we add proper locking around the `trades` field on the metrics handler, since we will face contention for it when writing from the `HandleFills` function which will be called from a separate thread.

nikhilsaraf (commented):
added comments to help with the general direction of this PR, thanks for the draft!

cmd/trade.go
debnil: If we're pre-initializing the handler (as per the suggestion [here](https://github.com/stellar/kelp/pull/574#discussion_r518378523)), then I don't think we need the registration at all.

plugins/tradeMetricsHandler.go
debnil: I think this is everything (including that VWAP formula) -- super helpful, thank you!! I've made these changes.

debnil:
@nikhilsaraf: awesome! I've done the bulk of this review, thanks so much. The only part that's left is the merging trade metrics with map. This only makes sense to do once the GUI metrics tracker is added in, since that changes the architecture of event updates.

plugins/tradeMetricsHandler.go
```diff
@@ -16,27 +16,19 @@ func MakeTradeMetricsHandler() *TradeMetricsHandler {
...
 
-// TotalBaseVolume returns the total base volume.
-func (h *TradeMetricsHandler) TotalBaseVolume() (total float64) {
-	for _, t := range h.trades {
-		total += t.Volume.AsFloat()
-	}
-	return
+// HandleFill handles a new trade
+// Implements FillHandler interface
+func (h *TradeMetricsHandler) HandleFill(trade model.Trade) error {
```
nikhilsaraf, line 31: we need to test this tradeMetricsHandler. For now let's have one simple test case, where we add 3 trades:
- buy XLM/USDT
- buy XLM/USDT
- sell XLM/USDT

then we assert that the computed values on the struct returned from `ComputeTradeMetrics` is correct.

this `TestComputeTradeMetrics` would be very simple and would ensure our calculations are correct.

nikhilsaraf (commented):
this is the right direction. I've added some inline comments and we can merge once completed.

we also need to add logic that will "start the trade metrics handler" process by registering it with the fill tracker.

you can take this PR out of draft stage and we can merge after the next set of changes.

support/metrics/metricsTracker.go
```diff
@@ -205,6 +207,7 @@ func MakeMetricsTracker(
 		botStartTime:        botStartTime,
 		isDisabled:          isDisabled,
 		updateEventSentTime: nil,
+		handler:             plugins.MakeTradeMetricsHandler(),
```
nikhilsaraf, line 210: @debnil we wouldn't construct this here -- it would get constructed outside here and passed in. the calling code would reuse this instance of tradeMetricsHandler to register it on the fill tracker which is where the HandleFill function will get invoked. does that make sense?

support/metrics/metricsTracker.go
```diff
@@ -312,40 +308,52 @@ func (mt *MetricsTracker) sendEvent(eventType string, eventProps interface{}) er
...
 	}
-	return
-}
 
-func (mt *MetricsTracker) getTotalNumTrades() (total int) {
-	for _, h := range mt.handlers {
-		total += h.NumTrades()
+	avgTradeSizeBase := totalBaseVolume / numTrades
+	avgTradeSizeQuote := totalQuoteVolume / numTrades
+	avgTradePrice := totalPrice / numTrades
```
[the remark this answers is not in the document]
nikhilsaraf: when we move this to the `tradeMetricsHandler` struct, the function can take in the `SecondsSinceLastUpdate` to help with this new computation.

support/metrics/metricsTracker.go
```diff
@@ -8,8 +8,7 @@ import (
 	"runtime/debug"
 	"time"
 
-	"github.com/stellar/kelp/api"
-	"github.com/stellar/kelp/model"
+	"github.com/stellar/kelp/plugins"
```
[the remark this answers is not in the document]
debnil: This makes sense, but as mentioned, it's a large change. I will address the rest of the comments.

I'm not sure what "make this change later" means in this context. Should I address this within this PR, but after the other outstanding PRs using the `metricsTracker` are merged? Or should we merge this as-is, and fix in a follow-up PR?

debnil:
@nikhilsaraf : most of the requested changes are made, but there's now an import cycle. `plugins/tradeMetricsHandler` imports `models.Trade`, which imports `plugins` indirectly. I don't think there's a way around the handler importing `models.Trade`, since that's the base data structure for all of its operations. Curious what you think the best fix is.

nikhilsaraf (commented):
can you point out where `model.Trade` imports `plugins`?

if there is a import cycle then the code won't compile.

PR #576, opened 2020-11-06
Clean up root binary invocation (closes #568)

### LINKED ISSUE #568: [1] clean up root.go basic kelp binary invocation logic
- take in isGUI param from command-line (or some other way of figuring out if it's the GUI bot or not)
- if isGUI, then default operation should be to load `kelp server` (windows loads with --no-electron)
- else loads the --help section, with the ASCII art
- remove the `.bat` file for windows since we don't need it any longer since we can execute the `kelp.exe` binary directly to load it in the browser

debnil:
This PR cleans up the root binary invocation logic, and closes #568.

### DIFF
--- a/cmd/server_amd64.go
+++ b/cmd/server_amd64.go
@@ -78,7 +78,7 @@ func init() {
 	options.horizonPubnetURI = serverCmd.Flags().String("horizon-pubnet-uri", "https://horizon.stellar.org", "URI to use for the horizon instance connected to the Stellar Public Network (must not contain the word 'test')")
 	options.noHeaders = serverCmd.Flags().Bool("no-headers", false, "do not use Amplitude or set X-App-Name and X-App-Version headers on requests to horizon")
 	options.verbose = serverCmd.Flags().BoolP("verbose", "v", false, "enable verbose log lines typically used for debugging")
-	options.noElectron = serverCmd.Flags().Bool("no-electron", false, "open in browser instead of using electron")
+	options.noElectron = serverCmd.Flags().Bool("no-electron", true, "open in browser instead of using electron")
 
 	serverCmd.Run = func(ccmd *cobra.Command, args []string) {
 		isLocalMode := env == envDev
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -409,10 +409,6 @@ do
         env GOOS=$GOOS GOARCH=$GOARCH GOARM=$GOARM go build -ldflags "$LDFLAGS" -o $ARCHIVE_DIR_SOURCE_UI/$GOOS-$GOARCH/kelp.exe
         check_build_result $?
         echo "successful"
-        
-        echo -n "copying over kelp-start.bat file to the windows build ..."
-        cp $KELP/gui/windows-bat-file/kelp-start.bat $ARCHIVE_DIR_SOURCE_UI/$GOOS-$GOARCH/
-        echo "done"
 
         # set paths needed for unzipping the vendor and ccxt files
         VENDOR_FILENAME=""


debnil:
@nikhilsaraf: marking this as a draft PR to make sure my approach is correct, and also because I'm not sure how to remove the `*.bat` file generated in the Windows build. Let me know if this makes sense thus far.

cmd/server_amd64.go
```diff
@@ -78,7 +78,7 @@ func init() {
 	options.horizonPubnetURI = serverCmd.Flags().String("horizon-pubnet-uri", "https://horizon.stellar.org", "URI to use for the horizon instance connected to the Stellar Public Network (must not contain the word 'test')")
 	options.noHeaders = serverCmd.Flags().Bool("no-headers", false, "do not use Amplitude or set X-App-Name and X-App-Version headers on requests to horizon")
 	options.verbose = serverCmd.Flags().BoolP("verbose", "v", false, "enable verbose log lines typically used for debugging")
-	options.noElectron = serverCmd.Flags().Bool("no-electron", false, "open in browser instead of using electron")
+	options.noElectron = serverCmd.Flags().Bool("no-electron", true, "open in browser instead of using electron")
```
nikhilsaraf, line 81: this should remain `false` as the default.

scripts/build.sh
```diff
@@ -409,10 +409,6 @@ do
         env GOOS=$GOOS GOARCH=$GOARCH GOARM=$GOARM go build -ldflags "$LDFLAGS" -o $ARCHIVE_DIR_SOURCE_UI/$GOOS-$GOARCH/kelp.exe
         check_build_result $?
         echo "successful"
-        
-        echo -n "copying over kelp-start.bat file to the windows build ..."
```
nikhilsaraf, line 413: we should also delete the `gui/windows-bat-file/kelp-start.bat` file in the repo along with this

cmd/server_amd64.go
nikhilsaraf: 1. the build script needs to set a *new* variable on the root.go that is true if the binary is built to be a GUI or not

- see the `hasUICapability` variable -- this is currently based on the architecture and included build files, we need something that is similar in usage but is manually injected from the build script
- there are two sections in the build script, one produces tar files for the CLI, and the other produces zip/dmg files (second half) for the GUI. the CLI binaries should say `isGUI=false`, and the GUI binary should say `isGUI=true`.

2. when starting up kelp in `root.go` we already split up execution based on the `hasUICapability` variable. Inside that we should check for GOOS and if it's windows we want to force set the `noElectron` field in the args to true before we start (update `args` variable before running `serverCmd.Run(ccmd, args)`)

debnil: Got it. Two questions: 

1. if we are checking inside the existing variable `hasUICapability`, where would we use the new `isGUI` variable?

2. Since `args` is an array of strings, what is the safest way to modify the argument corresponding to `noElectron`? A lot of approaches I'm seeing online assume a certain order of flags. Can we assume that with the `cobra` package?

nikhilsaraf: 1. for someone running on a windows amd64 system using the CLI, `hasUICapability` will be `true` but `isGUI` will be false. in this case we want the root command to `not` start up the GUI by default. this is why `isGUI` needs to be checked inside the case where `hasUICapability` is true.

2. In the case where it is windows and `isGUI` is true, maybe we can try this, you'll have to see if this will work: `options.noElectron = true` thereby overriding windows to run without electron regardless of what flag the user passed in on the command line. this way we won't need to mess with `args`

scripts/build.sh
nikhilsaraf: @debnil did we delete `gui/windows-bat-file/kelp-start.bat`?

nikhilsaraf:
committed as part of separate PR: https://github.com/stellar/kelp/pull/658

PR #580, opened 2020-11-10
Close popover menu on click (closes #301)

### LINKED ISSUE #301: clicking on the popover menu closes the menu
it's more usable when the menu closes after it's clicked

debnil:
This PR closes the popover menu on click, and closes #301.


debnil:
@nikhilsaraf: Let's go over this PR in our pair programming session quickly. I have zero idea why this isn't working and how to fix it. What happens is that when the options toggle is clicked, every `onClick` function for a component of the options is selected and run - so the two windows open, and the bot is deleted. I don't understand why that would happen, so I think there's some React concept I'm missing here.

PR #582, opened 2020-11-12
Add metrics to button (part of #508)

debnil:
This PR adds metrics functionality to `Button.js`. A subsequent PR will add the `eventName` to all invocations of `Button`.

### DIFF
--- a/gui/web/src/components/atoms/Button/Button.js
+++ b/gui/web/src/components/atoms/Button/Button.js
@@ -35,8 +35,6 @@ class Button extends Component {
     loading: false,
     onClick: () => {},
     disabled: false,
-    trackEvent: (eventName, values) => {},
-    eventName: '',    
   }
 
   static propTypes = {
@@ -46,7 +44,6 @@ class Button extends Component {
     onClick: PropTypes.func,
     loading: PropTypes.bool,
     disabled: PropTypes.bool,
-    trackEvent: PropTypes.func,
     eventName: PropTypes.string,
   };
   
@@ -55,9 +52,17 @@ class Button extends Component {
       return
     }
     
+    if (this.props.eventName === "" || this.props.eventName === undefined) {
+      return
+    }
+    
     var _this = this
-    // TODO DS populate the eventData.
-    this._asyncRequests["sendMetricEvent"] = sendMetricEvent(this.props.baseUrl, this.props.eventName, {}).then(resp => {
+    var eventData = {
+      'eventName': this.props.eventName,
+      'type': 'generic',
+      'component': 'button'
+    }
+    this._asyncRequests["sendMetricEvent"] = sendMetricEvent(this.props.baseUrl, this.props.eventName, eventData).then(resp => {
       if (!_this._asyncRequests["sendMetricEvent"]) {
         // if it has been deleted it means we don't want to process the result
         return
@@ -65,7 +70,9 @@ class Button extends Component {
       
       delete _this._asyncRequests["sendMetricEvent"];
       
-      // TODO DS Determine how to process resp
+      if (resp["success"] === undefined) {
+        console.log(resp["error"])
+      }
     })
   }
   
--- a/gui/web/src/components/atoms/Button/Button.js
+++ b/gui/web/src/components/atoms/Button/Button.js
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import styles from './Button.module.scss';
 import Icon from '../Icon/Icon';
 import LoadingAnimation from '../LoadingAnimation/LoadingAnimation';
+import sendMetricEvent from '../../../kelp-ops-api/sendMetricEvent';
 
 const iconSizes = {
   small:'10px',
@@ -18,6 +19,14 @@ const iconSizesRound = {
 }
 
 class Button extends Component {
+  constructor(props) {
+    super(props);
+    this.trackOnClick = this.trackOnClick.bind(this);
+    this.sendMetricEvent = this.sendMetricEvent.bind(this);
+    
+    this._asyncRequests = {};
+  }
+  
   static defaultProps = {
     icon: null,
     size: 'medium',
@@ -25,7 +34,9 @@ class Button extends Component {
     hsize: 'regular',
     loading: false,
     onClick: () => {},
-    disabled: false    
+    disabled: false,
+    trackEvent: (eventName, values) => {},
+    eventName: '',    
   }
 
   static propTypes = {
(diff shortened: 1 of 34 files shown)


debnil:
@nikhilsaraf Two questions for you:

1. What are the "values" that we'd send along with the metric?
2. I suspect that this `sendMetricEvent()` function will be used in many of the changes we are making. Where would be the best place to put it?

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -34,8 +45,34 @@ class Button extends Component {
...
+    disabled: PropTypes.bool,
+    trackEvent: PropTypes.func,
+    eventName: PropTypes.string,
   };
+  
+  sendMetricEvent() {
+    if (this._asyncRequests["sendMetricEvent"]) {
+      return
+    }
+
```
nikhilsaraf, line 57: we want to throw an error here if `eventName` is undefined or empty -- it should be defined for everyone and should not be empty.

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -18,14 +19,24 @@ const iconSizesRound = {
...
     icon: null,
     size: 'medium',
     variant: '',
     hsize: 'regular',
     loading: false,
     onClick: () => {},
-    disabled: false    
+    disabled: false,
+    trackEvent: (eventName, values) => {},
+    eventName: '',
```
nikhilsaraf, line 39: no default value for eventName

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -34,8 +45,34 @@ class Button extends Component {
...
+    // TODO DS populate the eventData.
+    this._asyncRequests["sendMetricEvent"] = sendMetricEvent(this.props.baseUrl, this.props.eventName, {}).then(resp => {
+      if (!_this._asyncRequests["sendMetricEvent"]) {
+        // if it has been deleted it means we don't want to process the result
+        return
+      }
+      
+      delete _this._asyncRequests["sendMetricEvent"];
+      
+      // TODO DS Determine how to process resp
```
nikhilsaraf, line 68: if resp is success then do nothing, if there's an error then just console.log it for now.

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -34,8 +45,34 @@ class Button extends Component {
     variant: PropTypes.string,
     onClick: PropTypes.func,
     loading: PropTypes.bool,
-    disabled: PropTypes.bool
+    disabled: PropTypes.bool,
+    trackEvent: PropTypes.func,
```
nikhilsaraf, line 49: what is `trackEvent` -- doesn't seem like it's used anywhere?

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -34,8 +45,34 @@ class Button extends Component {
...
+    eventName: PropTypes.string,
   };
+  
+  sendMetricEvent() {
+    if (this._asyncRequests["sendMetricEvent"]) {
+      return
+    }
+    
+    var _this = this
+    // TODO DS populate the eventData.
```
nikhilsaraf, line 59:
```
{
    'eventName': eventName,
    'type': 'generic',
    'component': 'button'
}
```

please add any more info that is relevant to the data

nikhilsaraf (commented):
can mark as non-draft so we can merge in the next round.

gui/web/src/components/atoms/Button/Button.js
debnil: Honestly, since it's been a few weeks, I just don't remember. I think this would be the function that wraps the event in some format, but I just can't remember. I'll delete this for now. If it matters, we'll find out, and no downside since this won't break anything. How does that sound?

nikhilsaraf: perfect.

gui/web/src/components/atoms/Button/Button.js
nikhilsaraf: maybe the right thing to do is to let React handle part of this for us.
i.e. the `eventName` prop should be `PropTypes.string.isRequired` and then we only need to check if this is an empty string here. if it's an empty string we should console.log an error before returning.

if React can handle the validation for us as well, something like this (made up my syntax here):
`PropTypes.string.isRequired(exclude=[""])` then that would be great. IIRC React has some validation functionality like this, but don't spend too much time on trying to figure it out if it's not straightforward.

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -55,17 +52,27 @@ class Button extends Component {
...
     }
     
+    if (this.props.eventName === "" || this.props.eventName === undefined) {
+      return
+    }
+    
     var _this = this
-    // TODO DS populate the eventData.
-    this._asyncRequests["sendMetricEvent"] = sendMetricEvent(this.props.baseUrl, this.props.eventName, {}).then(resp => {
+    var eventData = {
```
nikhilsaraf, line 60: `const` instead of `var` (I should have been more diligent about this earlier to have set a better precedent)

nikhilsaraf: PS: we can probably remove the single-quotes around the field, since this is javascript in which json is native

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -55,17 +52,27 @@ class Button extends Component {
...
+    this._asyncRequests["sendMetricEvent"] = sendMetricEvent(this.props.baseUrl, this.props.eventName, eventData).then(resp => {
       if (!_this._asyncRequests["sendMetricEvent"]) {
         // if it has been deleted it means we don't want to process the result
         return
       }
       
       delete _this._asyncRequests["sendMetricEvent"];
       
-      // TODO DS Determine how to process resp
+      if (resp["success"] === undefined) {
```
nikhilsaraf, line 73: use resp.success and resp.error instead of square bracket approach

specifically, here we should use `if (!resp.success)` so it checks for undefined and null. I think that should work, but will need to be tested. you can hardcode returning a failure on the backend to see if it logs. please add screenshot of the test. this can server as a substitute for an automated test for now.

gui/web/src/components/atoms/Button/Button.js
```diff
@@ -55,17 +52,27 @@ class Button extends Component {
...
+        console.log(resp["error"])
+      }
```
nikhilsaraf, line 75: add a comment like this after this so we know we have explicitly chosen to ignore the else case:
`// else do nothing on success`

nikhilsaraf (commented):
seems like should be ready in the next round.

gui/web/src/components/atoms/Button/Button.js
nikhilsaraf: handled in new PR I'm putting together

nikhilsaraf:
closing because work included in https://github.com/stellar/kelp/pull/642

PR #586, opened 2020-11-13
Add buy infrastructure to volume filter (part of #522)

debnil:
This PR adds the buy infrastructure to the volume filter, as per step 2 [here](https://github.com/stellar/kelp/pull/548#issuecomment-712792629).


plugins/volumeFilter.go
```diff
@@ -179,50 +180,50 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
...
 			// update the dailyTBB to include the additional amounts so they can be used in the calculation of the next operation
-			*dailyTBBAccumulator.SellBaseAssetCapInBaseUnits += newAmountBeingSold
-			*dailyTBBAccumulator.SellBaseAssetCapInQuoteUnits += (newAmountBeingSold * sellPrice)
+			*dailyTBBAccumulator.BaseAssetCapInBaseUnits += newAmountBeingSold
+			*dailyTBBAccumulator.BaseAssetCapInQuoteUnits += (newAmountBeingSold * sellPrice)
 			return opToReturn, nil
 		}
 	} else {
-		// TODO buying side
+		// TODO buying side - we need to implement this to support buy side filters; extract common logic from the above sell side case
```
nikhilsaraf, line 226: @debnil I think we are doing this PR too early. the next step is to fill in this function, and add the relevant tests to the buyVolume filter test (and any other tests related to the buy side).

once that's done, the refactor should be the next step. this will not be an exact cherry-pick from the previous set of commits I had added because of some of the volume filter changes we made.

as a last step we should add a new buy twap strategy with the config file. doing it in steps will make for a smaller PR.

nikhilsaraf (commented):
as commented inline, I think we should hold off on this PR

plugins/volumeFilter_test.go
```diff
@@ -100,48 +100,54 @@ func TestMakeFilterVolume(t *testing.T) {
...
-						k.exchangeName,
-						tradingPair,
-						testAssetDisplayFn,
-						utils.NativeAsset,
-						utils.NativeAsset,
-						&sql.DB{},
-						config,
-					)
+					// TODO DS Vary filter action between buy and sell, once buy logic is implemented.
+					wantFilter := makeWantVolumeFilter(config, k.wantMarketIDs, k.accountIDs, queries.DailyVolumeActionSell)
```
nikhilsaraf, line 131: this should be `action` instead of hardcoding to `DailyVolumeActionSell`

queries/dailyVolumeByDate_test.go
```diff
@@ -47,7 +48,18 @@ func TestDailyVolumeByDate_QueryRow(t *testing.T) {
 		wantTomorrowBase          float64
 		wantTomorrowQuote         float64
 	}{
+		// TODO add case for buy base/quote and add trade data in test below accordingly
```
nikhilsaraf, line 51: please add this case in a subsequent PR (better to separate out so we're not complicating this PR)

queries/dailyVolumeByDate_test.go
```diff
@@ -98,113 +115,113 @@ func TestDailyVolumeByDate_QueryRow(t *testing.T) {
...
-			db := connectTestDb()
-			defer db.Close()
-			for _, s := range setupStatements {
-				_, e := db.Exec(s)
-				if e != nil {
-					panic(e)
-				}
-			}
+	// setup db
+	yesterday, _ := time.Parse(time.RFC3339, "2020-01-20T15:00:00Z")
```
nikhilsaraf, line 119: was this a direct copy paste or did something change inside this too?

nikhilsaraf (commented):
few minor comments, please fix those up and we should be good to merge.
can you leave the comments unresolved so I can check each item that is fixed up? will be a bit easier for me to track.

any leftover items, please address in a new PR

queries/dailyVolumeByDate_test.go
debnil: That was the plan - the TODO was so I'd remember this in the next step of implementation. Should I not do that?

queries/dailyVolumeByDate_test.go
debnil: Direct copy paste - nothing changed, just a small performance fix. 

Previously, the tests were setting up a brand new database in each iteration of the for-loop / for each test case. I found that this was making the run-time substantially slower. I moved the setup steps out of the loop and made it about 5x faster. (.124s vs .6s). I don't think this is happening on the CI, but it's a solid saving without losing any correctness.

plugins/volumeFilter.go
debnil: We discussed this earlier today, but for documentation, this is the right stage for this fix. We can't implement this function absent the other infrastructural changes in this PR.

plugins/filterFactory_test.go
```diff
@@ -135,50 +136,71 @@ func TestMakeVolumeFilterConfig(t *testing.T) {
...
 		{
 			configInput: "volume/daily/sell/base/3500.0/exact",
 			wantConfig: &VolumeFilterConfig{
-				SellBaseAssetCapInBaseUnits:  pointy.Float64(3500.0),
-				SellBaseAssetCapInQuoteUnits: nil,
-				mode:                         volumeFilterModeExact,
-				additionalMarketIDs:          nil,
-				optionalAccountIDs:           nil,
+				BaseAssetCapInBaseUnits:  pointy.Float64(3500.0),
+				BaseAssetCapInQuoteUnits: nil,
```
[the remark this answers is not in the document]
debnil: Sounds good. Since this is updating an old test, I figured that like our more recent tests, we'd want to vary both mode and action. To do this more efficiently, I changed the `configInput` to a format string and injected both mode and action. If the previous way would be preferred, let me know and I'm happy to change it back.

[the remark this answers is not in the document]
nikhilsaraf: the previous way is a little better because it's more readable, a direct mapping from input to output.

plugins/volumeFilter_test.go
```diff
@@ -287,36 +287,37 @@ func TestVolumeFilterFn(t *testing.T) {
...
-			}
-
-			if k.sellBaseCapInBase != nil && k.sellBaseCapInQuote != nil {
-				assert.Fail(t, "both of the cap values cannot be set")
-				return
-			}
+		for _, action := range []queries.DailyVolumeAction{queries.DailyVolumeActionSell} {
+			t.Run(k.name, func(t *testing.T) {
+				// exactly one of the two cap values must be set
+				if k.sellBaseCapInBase == nil && k.sellBaseCapInQuote == nil {
```
nikhilsaraf, line 293: these should be called `baseCapInBase` and `baseCapInQuote` instead of `sellBaseCapInBase` and `sellBaseCapInQuote`. Better to do this in a separate PR since the overhead of this PR is already getting complex. It's ok to do this when you add the `buy` action to this for loop.

plugins/volumeFilter_test.go
```diff
@@ -287,36 +287,37 @@ func TestVolumeFilterFn(t *testing.T) {
...
-			if k.sellBaseCapInBase == nil && k.sellBaseCapInQuote == nil {
-				assert.Fail(t, "either one of the two cap values must be set")
-				return
-			}
-
-			if k.sellBaseCapInBase != nil && k.sellBaseCapInQuote != nil {
-				assert.Fail(t, "both of the cap values cannot be set")
-				return
-			}
+		for _, action := range []queries.DailyVolumeAction{queries.DailyVolumeActionSell} {
```
nikhilsaraf, line 290: PR submission tip:
it's best not to add changes like this in follow-on PR changes since it introduces a large amount of red and green lines. As a reviewer, this makes reviewing this PR 10x more complex, since I have to ensure nothing has changed in here and whatever has changed makes sense, etc.
I say this now, because I'm hoping to start leaving similar tips on PRs that will allow us to increase our PR efficiency (such as the comment about not resolving review comments). Hopefully this will reduce the PR fatigue for both of us, and happy to get similar thoughts and tips from you as well :)

queries/dailyVolumeByDate_test.go
nikhilsaraf: yes, good fix! 👍

plugins/filterFactory_test.go
```diff
@@ -200,8 +229,8 @@ func assertVolumeFilterConfigEqual(t *testing.T, want *VolumeFilterConfig, actua
 	} else if actual == nil {
 		assert.Fail(t, fmt.Sprintf("actual was nil but expected %v", *want))
 	} else {
-		assert.Equal(t, want.SellBaseAssetCapInBaseUnits, actual.SellBaseAssetCapInBaseUnits)
-		assert.Equal(t, want.SellBaseAssetCapInQuoteUnits, actual.SellBaseAssetCapInQuoteUnits)
+		assert.Equal(t, want.BaseAssetCapInBaseUnits, actual.BaseAssetCapInBaseUnits)
+		assert.Equal(t, want.BaseAssetCapInQuoteUnits, actual.BaseAssetCapInQuoteUnits)
```
nikhilsaraf, line 233: @debnil add a check on the `action` field here

nikhilsaraf (commented):
I think there's only 1 change below and we should be good to merge after that.
Any follow-on items can be done in a subsequent PR. thanks! 👍

nikhilsaraf (approved):
lgtm, thanks 🚢

PR #589, opened 2020-11-20
Validate volume filter config (closes #571)

### LINKED ISSUE #571: [ad-hoc] [1] Add validation to makeFilterVolume
We want to add validation to the config (of type `*VolumeFilterConfig` in `makeFilterVolume` in `plugins/volumeFilter.go`. This validation should include checking that exactly one asset cap is non-nil; the mode is valid; and optional accountIDs and market IDs are both non-nil.

Once the above is added, a test case should be added to `TestMakeFilterVolume` within `volumeFilter_test.go`. This test case should confirm that if an invalid configuration is constructed, an error is thrown by `makeFilterVolume`.

debnil:
This PR adds additional validation for the volume filter config; calls that validation within the volume filter constructor; and adds necessary tests for that validation. It also edits other tests of the volume config that now fail because of validation. This closes #571.

### DIFF
--- a/plugins/filterFactory.go
+++ b/plugins/filterFactory.go
@@ -9,6 +9,7 @@ import (
 
 	hProtocol "github.com/stellar/go/protocols/horizon"
 	"github.com/stellar/kelp/model"
+	"github.com/stellar/kelp/queries"
 )
 
 var filterIDRegex *regexp.Regexp
@@ -72,18 +73,20 @@ func filterVolume(f *FilterFactory, configInput string) (SubmitFilter, error) {
 }
 
 func makeRawVolumeFilterConfig(
-	sellBaseAssetCapInBaseUnits *float64,
-	sellBaseAssetCapInQuoteUnits *float64,
+	baseAssetCapInBaseUnits *float64,
+	baseAssetCapInQuoteUnits *float64,
+	action queries.DailyVolumeAction,
 	mode volumeFilterMode,
 	additionalMarketIDs []string,
 	optionalAccountIDs []string,
 ) *VolumeFilterConfig {
 	return &VolumeFilterConfig{
-		SellBaseAssetCapInBaseUnits:  sellBaseAssetCapInBaseUnits,
-		SellBaseAssetCapInQuoteUnits: sellBaseAssetCapInQuoteUnits,
-		mode:                         mode,
-		additionalMarketIDs:          additionalMarketIDs,
-		optionalAccountIDs:           optionalAccountIDs,
+		BaseAssetCapInBaseUnits:  baseAssetCapInBaseUnits,
+		BaseAssetCapInQuoteUnits: baseAssetCapInQuoteUnits,
+		action:                   action,
+		mode:                     mode,
+		additionalMarketIDs:      additionalMarketIDs,
+		optionalAccountIDs:       optionalAccountIDs,
 	}
 }
 
@@ -103,6 +106,12 @@ func makeVolumeFilterConfig(configInput string) (*VolumeFilterConfig, error) {
 	if limitWindowParts[0] != "daily" {
 		return nil, fmt.Errorf("invalid input (%s), the second part needs to equal or start with \"daily\"", configInput)
 	}
+
+	action, e := queries.ParseDailyVolumeAction(parts[2])
+	if e != nil {
+		return nil, fmt.Errorf("could not parse volume filter action from input (%s): %s", configInput, e)
+	}
+	config.action = action
 
 	errInvalid := fmt.Errorf("invalid input (%s), the modifier for \"daily\" can be either \"market_ids\" or \"account_ids\" like so 'daily:market_ids=[4c19915f47,db4531d586]' or 'daily:account_ids=[account1,account2]' or 'daily:market_ids=[4c19915f47,db4531d586]:account_ids=[account1,account2]'", configInput)
 	if len(limitWindowParts) == 2 {
@@ -124,17 +133,14 @@ func makeVolumeFilterConfig(configInput string) (*VolumeFilterConfig, error) {
 		return nil, fmt.Errorf("invalid input (%s), the second part needs to be \"daily\" and can have only one modifier \"market_ids\" like so 'daily:market_ids=[4c19915f47,db4531d586]'", configInput)
 	}
 
-	if parts[2] != "sell" {
-		return nil, fmt.Errorf("invalid input (%s), the third part needs to be \"sell\"", configInput)
-	}
 	limit, e := strconv.ParseFloat(parts[4], 64)
 	if e != nil {
 		return nil, fmt.Errorf("could not parse the fourth part as a float value from config value (%s): %s", configInput, e)
 	}
 	if parts[3] == "base" {
-		config.SellBaseAssetCapInBaseUnits = &limit
+		config.BaseAssetCapInBaseUnits = &limit
 	} else if parts[3] == "quote" {
-		config.SellBaseAssetCapInQuoteUnits = &limit
+		config.BaseAssetCapInQuoteUnits = &limit
 	} else {
 		return nil, fmt.Errorf("invalid input (%s), the third part needs to be \"base\" or \"quote\"", configInput)
 	}
--- a/plugins/filterFactory.go
+++ b/plugins/filterFactory.go
@@ -145,6 +145,15 @@ func makeVolumeFilterConfig(configInput string) (*VolumeFilterConfig, error) {
 		return nil, fmt.Errorf("invalid input (%s), the third part needs to be \"base\" or \"quote\"", configInput)
 	}
 
+	// marketIDs and account IDs might not be present in the config string, but cannot be nil on the config
+	if config.additionalMarketIDs == nil {
+		config.additionalMarketIDs = []string{}
+	}
+
+	if config.optionalAccountIDs == nil {
+		config.optionalAccountIDs = []string{}
+	}
+
 	if e = config.Validate(); e != nil {
 		return nil, fmt.Errorf("invalid input (%s), did not pass validation: %s", configInput, e)
 	}
--- a/plugins/filterFactory.go
+++ b/plugins/filterFactory.go
@@ -145,15 +145,6 @@ func makeVolumeFilterConfig(configInput string) (*VolumeFilterConfig, error) {
 		return nil, fmt.Errorf("invalid input (%s), the third part needs to be \"base\" or \"quote\"", configInput)
 	}
 
-	// marketIDs and account IDs might not be present in the config string, but cannot be nil on the config
-	if config.additionalMarketIDs == nil {
-		config.additionalMarketIDs = []string{}
-	}
-
-	if config.optionalAccountIDs == nil {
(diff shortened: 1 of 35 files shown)


plugins/volumeFilter.go
```diff
@@ -103,8 +106,25 @@ var _ SubmitFilter = &volumeFilter{}
...
+		return fmt.Errorf("invalid asset caps: only one asset cap can be non-nil, but both are non-nil")
+	}
+
+	if c.SellBaseAssetCapInBaseUnits == nil && c.SellBaseAssetCapInQuoteUnits == nil {
+		return fmt.Errorf("invalid asset caps: only one asset cap can be non-nil, but both are nil")
+	}
+
+	// return the original error, as it is already well-formatted
+	if _, e := parseVolumeFilterMode(string(c.mode)); e != nil {
+		return e
```
nikhilsaraf, line 119: we should always wrap the error as a habit as it helps trace the error path. long term we will need to find a less verbose solution, but this has been working well so far.

plugins/volumeFilter.go
```diff
@@ -103,8 +106,25 @@ var _ SubmitFilter = &volumeFilter{}
 
 // Validate ensures validity
 func (c *VolumeFilterConfig) Validate() error {
-	if c.isEmpty() {
-		return fmt.Errorf("the volumeFilterConfig was empty")
+	if c.SellBaseAssetCapInBaseUnits != nil && c.SellBaseAssetCapInQuoteUnits != nil {
```
nikhilsaraf, line 109: drop the `Sell` prefix, like we discussed in slack

plugins/volumeFilter_test.go
```diff
@@ -322,3 +321,87 @@ func makeManageSellOffer(price string, amount string) *txnbuild.ManageSellOffer
 		Amount:  amount,
 	}
 }
+
+func TestValidateConfig(t *testing.T) {
+	testCases := []struct {
+		name             string
+		sellBaseCapBase  *float64
+		sellBaseCapQuote *float64
+		mode             volumeFilterMode
```
nikhilsaraf, line 330: should take action as an input too.

don't need to add more than 1 test case to make sure it fails on an invalid value.
all existing cases should take a combination of `buy` and `sell` action (or maybe all are `sell` but only one success case is `buy` to prove the buy case also works)

plugins/volumeFilter.go
```diff
@@ -103,8 +106,25 @@ var _ SubmitFilter = &volumeFilter{}
...
+
+	if c.SellBaseAssetCapInBaseUnits == nil && c.SellBaseAssetCapInQuoteUnits == nil {
+		return fmt.Errorf("invalid asset caps: only one asset cap can be non-nil, but both are nil")
+	}
+
+	// return the original error, as it is already well-formatted
+	if _, e := parseVolumeFilterMode(string(c.mode)); e != nil {
+		return e
+	}
+
```
nikhilsaraf, line 121: needs to validate `action` as well

plugins/volumeFilter_test.go
```diff
@@ -322,3 +321,87 @@ func makeManageSellOffer(price string, amount string) *txnbuild.ManageSellOffer
...
+			accountIDs:       nil,
+			wantErr:          fmt.Errorf("invalid optional account ids: must be non-nil"),
+		},
+	}
+
+	for _, k := range testCases {
+		t.Run(k.name, func(t *testing.T) {
+			c := makeRawVolumeFilterConfig(k.sellBaseCapBase, k.sellBaseCapQuote, k.mode, k.marketIDs, k.accountIDs)
+			gotErr := c.Validate()
+			assert.Equal(t, gotErr, k.wantErr)
```
nikhilsaraf, line 404: should `wantErr` be the second arg and `gotErr` the third arg?

nikhilsaraf (commented):
almost done

plugins/volumeFilter.go
debnil: Yup - those changes have all been made with the buy-side implementation. I'll merge that in and then update this branch.

plugins/volumeFilter_test.go
debnil: Makes sense. To make this more efficient (vary across buy and sell for existing tests), I've just done the same loop over modes and actions we've done in other test cases.

plugins/filterFactory.go
```diff
@@ -145,6 +145,15 @@ func makeVolumeFilterConfig(configInput string) (*VolumeFilterConfig, error) {
 		return nil, fmt.Errorf("invalid input (%s), the third part needs to be \"base\" or \"quote\"", configInput)
 	}
 
+	// marketIDs and account IDs might not be present in the config string, but cannot be nil on the config
+	if config.additionalMarketIDs == nil {
```
nikhilsaraf, line 149: @debnil this is significantly reducing the quality of the codebase. now we have two places that "set" these two arrays, here and in `parseVolumeFilterModifier`. please remove.

I'm not sure if we had discussed adding this logic, but if there was a previous comment that indicated it please let me know so I can clarify requested changes better.

plugins/filterFactory_test.go
```diff
@@ -143,48 +143,48 @@ func TestMakeVolumeFilterConfig(t *testing.T) {
 			wantConfig: &VolumeFilterConfig{
 				BaseAssetCapInBaseUnits:  pointy.Float64(3500.0),
 				BaseAssetCapInQuoteUnits: nil,
-				additionalMarketIDs:      nil,
-				optionalAccountIDs:       nil,
+				additionalMarketIDs:      []string{},
```
nikhilsaraf, line 146: don't recall requesting this change in my previous set of comments. why did these lines change?

plugins/volumeFilter_test.go
nikhilsaraf: the test has become more complex with this change, therefore harder to maintain.
there's also too many red and green lines which is increasing the PR overhead.
please revert and do the changes that were requested above with only an addition of 1 test case.

nikhilsaraf (changes requested):
I thought we were almost done in the last review, but there's a lot of red and green lines from the last iteration of the review. Inline I've requested to revert most of the "new ideas" and to fix up the suggestions made in the previous iteration.

See the new [Pull Request Guide](https://github.com/stellar/kelp/blob/master/.github/.pull_request_template.md) for details on the expectations when submitting Pull Requests.

plugins/filterFactory.go
debnil: These lines changed because now, when we validate the config, we check that market IDs and account IDs are not nil. This function is what creates a config. Previously, when parsing a config string, if it did not specify

The point of this pull request is to validate the config. That requires making changes like this to other functions that are now invalid. We could in theory split these in a separate PR, but that seems meaningless without the validation changes. If these changes are a bad idea, then we should either reconsider the validation scheme or change how the parsing logic works above this - I think this approach does that in the cleanest way possible, given the way that market and account IDs are parsed.

plugins/filterFactory_test.go
debnil: I sent a message on Slack about this when I sent the PR, but you're right that it should get posted on the PR too. Adding it here:

I had to make a couple changes in other tests that are now failing due to validation. It doesn't really make sense to separate these changes into other PRs, since the changes matter because of validation. These are things like "no nil marketIDs or optionalAccountIDs on volumeFilterConfig after parsing".

plugins/volumeFilter_test.go
```diff
@@ -329,3 +328,84 @@ func makeManageSellOffer(price string, amount string) *txnbuild.ManageSellOffer
...
+			name:         "failure - nil account IDs",
+			baseCapBase:  pointy.Float64(1.0),
+			baseCapQuote: nil,
+			marketIDs:    []string{},
+			accountIDs:   nil,
+			wantErr:      fmt.Errorf("invalid optional account ids: must be non-nil"),
+		},
+	}
+
+	for _, mode := range []volumeFilterMode{volumeFilterModeExact, volumeFilterModeIgnore} {
```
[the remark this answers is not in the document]
debnil: Those changes are already done. It now loops over both buy and sell actions (which is what these comments request). I'm also looping over the mode, as that's the testing pattern we've used in the tests done in the last few weeks - as I put out this PR 2-3 weeks ago, before we started varying over modes and actions in these config tests. But the mode loop is not essential, and I can hard code the mode in the testing struct if you think it makes sense.

I've also added two invalid test cases - one to test invalid action, one to test invalid mode, since those wouldn't be caught by the above testing. 2 versus 1 failure cases isn't a huge difference, especially when both test otherwise untested failure cases.

That said, this is making me reconsider the quantitative checks of red and green lines outlined in the PR policy we're discussing. The vast majority of the red and green lines in this file are because of a level of indentation changing. This is directly because of the request to test buy and sell actions for most test cases. It's just easier to do that systematically than pick some successful cases and create a couple of test cases. The additional changes due to mode are (+1, -10): one additional line to loop over the mode, and ten deletions of the mode from each test. 

As discussed above, the other changes displayed are the result of separate changes that came up when I ran these tests locally, found a bunch of tests now failing because of validation, and fixed those. I'm not sure why the CI didn't trigger those failures, since validate was being called before. But I don't think not including them in this PR is an option. Agreed the process level fix here is to include a comment about this - which I sent over Slack but somehow didn't get commented to the PR - that's on me, so all good there.

plugins/filterFactory_test.go
debnil: Oh actually, I did notice these CI failures - that's why I didn't request a review on this two weeks ago - was planning on coming back and fixing those after the other PRs in buy TWAP were done, you just beat me to the review :) The version that you reviewed (the initial commit) has all of these failures.

plugins/volumeFilter.go
```diff
@@ -102,8 +105,28 @@ var _ SubmitFilter = &volumeFilter{}
...
+	if _, e := parseVolumeFilterMode(string(c.mode)); e != nil {
+		return fmt.Errorf("could not parse mode: %s", e)
+	}
+
+	if _, e := queries.ParseDailyVolumeAction(string(c.action)); e != nil {
+		return fmt.Errorf("could not parse action: %s", e)
+	}
+
+	if c.additionalMarketIDs == nil {
```
nikhilsaraf, line 124: remove these `nil` errors. I think this is what is causing all the issues. these are not required since `nil` is a valid value for both arrays.

nikhilsaraf (commented):
I think we have a very clean way forward now. let's try and land this in the next shot.

plugins/volumeFilter.go
debnil: Got it, this makes sense and completely resolves the issue.

plugins/volumeFilter_test.go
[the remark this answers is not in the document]
debnil: Yup, perfect.

plugins/filterFactory.go
debnil: Sounds good - these are good clarifications, changes made!

nikhilsaraf (approved):
this looks great, thank you @debnil ! 🎉

PR #600, opened 2020-12-02
Rename caps in volume filter tests (part of #522)

debnil:
This PR renames caps in the volume filter tests, as part of #522. This renames the various fields called `SellBaseAssetCap*` as `BaseAssetCap*`. This reduces the bloat for follow-up PRs, which will add functional changes and additional tests.

### DIFF
--- a/plugins/volumeFilter.go
+++ b/plugins/volumeFilter.go
@@ -37,6 +37,7 @@ func parseVolumeFilterMode(mode string) (volumeFilterMode, error) {
 type VolumeFilterConfig struct {
 	BaseAssetCapInBaseUnits  *float64
 	BaseAssetCapInQuoteUnits *float64
+	action                   queries.DailyVolumeAction
 	mode                     volumeFilterMode
 	additionalMarketIDs      []string
 	optionalAccountIDs       []string
@@ -80,7 +81,7 @@ func makeFilterVolume(
 
 	marketID := MakeMarketID(exchangeName, baseAssetString, quoteAssetString)
 	marketIDs := utils.Dedupe(append([]string{marketID}, config.additionalMarketIDs...))
-	dailyVolumeByDateQuery, e := queries.MakeDailyVolumeByDateForMarketIdsAction(db, marketIDs, "sell", config.optionalAccountIDs)
+	dailyVolumeByDateQuery, e := queries.MakeDailyVolumeByDateForMarketIdsAction(db, marketIDs, config.action, config.optionalAccountIDs)
 	if e != nil {
 		return nil, fmt.Errorf("could not make daily volume by date Query: %s", e)
 	}
@@ -134,10 +135,10 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
 		BaseAssetCapInQuoteUnits: &dailyValuesBaseSold.QuoteVol,
 	}
 	// daily to-be-booked starts out as empty and accumulates the values of the operations
-	dailyTbbSellBase := 0.0
+	dailyTbbBase := 0.0
 	dailyTbbSellQuote := 0.0
 	dailyTBB := &VolumeFilterConfig{
-		BaseAssetCapInBaseUnits:  &dailyTbbSellBase,
+		BaseAssetCapInBaseUnits:  &dailyTbbBase,
 		BaseAssetCapInQuoteUnits: &dailyTbbSellQuote,
 	}
 
@@ -285,7 +286,7 @@ func (f *volumeFilter) isSellingBase() bool {
 func (f *volumeFilter) mustGetBaseAssetCapInBaseUnits() (float64, error) {
 	value := f.config.BaseAssetCapInBaseUnits
 	if value == nil {
-		return 0.0, fmt.Errorf("baseAssetCapInBaseUnits is nil, config = %v", f.config)
+		return 0.0, fmt.Errorf("BaseAssetCapInBaseUnits is nil, config = %v", f.config)
 	}
 	return *value, nil
 }
--- a/plugins/volumeFilter.go
+++ b/plugins/volumeFilter.go
@@ -110,8 +110,8 @@ func (c *VolumeFilterConfig) Validate() error {
 
 // String is the stringer method
 func (c *VolumeFilterConfig) String() string {
-	return fmt.Sprintf("VolumeFilterConfig[BaseAssetCapInBaseUnits=%s, BaseAssetCapInQuoteUnits=%s, mode=%s, additionalMarketIDs=%v, optionalAccountIDs=%v]",
-		utils.CheckedFloatPtr(c.BaseAssetCapInBaseUnits), utils.CheckedFloatPtr(c.BaseAssetCapInQuoteUnits), c.mode, c.additionalMarketIDs, c.optionalAccountIDs)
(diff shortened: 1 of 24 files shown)


plugins/volumeFilter.go
```diff
@@ -110,8 +110,8 @@ func (c *VolumeFilterConfig) Validate() error {
 
 // String is the stringer method
 func (c *VolumeFilterConfig) String() string {
-	return fmt.Sprintf("VolumeFilterConfig[BaseAssetCapInBaseUnits=%s, BaseAssetCapInQuoteUnits=%s, mode=%s, action=%s, additionalMarketIDs=%v, optionalAccountIDs=%v]",
-		utils.CheckedFloatPtr(c.BaseAssetCapInBaseUnits), utils.CheckedFloatPtr(c.BaseAssetCapInQuoteUnits), c.mode, c.action, c.additionalMarketIDs, c.optionalAccountIDs)
+	return fmt.Sprintf("VolumeFilterConfig[baseAssetCapInBaseUnits=%s, baseAssetCapInQuoteUnits=%s, mode=%s, additionalMarketIDs=%v, optionalAccountIDs=%v]",
```
nikhilsaraf, line 113: `baseAssetCapInBaseUnits` should remain as `BaseAssetCapInBaseUnits` as well as for the other field.

these names should follow the variable names. this can be autogenerated at some point, but we can continue to do manually for now.

nikhilsaraf: PS: once we eliminate all direct calls to construct VolumeFilterConfig once it all goes through the raw factory method (a pending TODO), then we can un-export these two fields.

plugins/volumeFilter.go
```diff
@@ -163,23 +163,23 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
...
 	}
 
-	sellPrice, e := strconv.ParseFloat(op.Price, 64)
+	offerPrice, e := strconv.ParseFloat(op.Price, 64)
 	if e != nil {
 		return nil, fmt.Errorf("could not convert price (%s) to float: %s", op.Price, e)
 	}
 
-	amountValueUnitsBeingSold, e := strconv.ParseFloat(op.Amount, 64)
+	offerAmountValueUnits, e := strconv.ParseFloat(op.Amount, 64)
```
nikhilsaraf, line 171: @debnil this variable name change has introduced unwanted ambiguity..

`amountValueUnitsBeingSold` tells us that the value corresponds to the selling asset in the manageSellOffer.
`offerAmountValueUnits` does not tell us whether the amount corresponds to the selling asset or buying asset in the manageSellOffer.

1. need to figure out what this variable means when we use `buy` types, which is not super-straightforward -- may need to take a TDD approach or intrusive debugging to figure this out.
2. need to name this variable accordingly so it is super clear.

the reason we need to do the two steps above is because we unfortunately had solved the trivial case previously -- doing a "sell" type operation using a manageSellOffer where amount and price all follow the selling asset (because that's how manageSellOffer works). the hard part is to now figure out what values we will see when we invert it.

let's discuss in our 1-1 session tomorrow (Thursday) about what you end up finding from your debugging.

plugins/volumeFilter.go
```diff
@@ -191,37 +191,82 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
...
 		if keepSellingBase && keepSellingQuote {
 			// update the dailyTBB to include the additional amounts so they can be used in the calculation of the next operation
 			*dailyTBBAccumulator.BaseAssetCapInBaseUnits += newAmountBeingSold
-			*dailyTBBAccumulator.BaseAssetCapInQuoteUnits += (newAmountBeingSold * sellPrice)
+			*dailyTBBAccumulator.BaseAssetCapInQuoteUnits += (newAmountBeingSold * offerPrice)
 			return opToReturn, nil
 		}
 	} else {
-		// TODO buying side - we need to implement this to support buy side filters; extract common logic from the above sell side case
+		opToReturn := op
```
nikhilsaraf, line 224: As you might have guessed, this is the main meat of the buytwap logic. Everything else we've done so far was basically just plumbing work to get us here.

I think the right thing to do is to extract the common logic from the above sell side case to simplify this. For example, things like `projectedBoughtInBaseUnits` should logically represent `projectedExposureInBaseUnits` instead so it can be reused for both buy and sell.

this basically boils down to the fact that selling `XLM/USD` is the same as buying `USD/XLM` so we don't need any if conditions or new logic, but we need some sort of "translation layer" that converts both the buy/sell forms to a more general form where we can phrase it as "exposure" to a specific asset (where exposure means the likely consumption of that asset from the balance because we give it up) vs explicitly saying amount is being bought/sold. The end result should make it "immediately obvious" what is happening in the code.

To do this it may be helpful to grok exactly what is going on in the above sell case and refactor the logic to be more generalized to both buy and sell cases. In the past, I've had success with modeling the code blocks as a _flowchart_ with distinct logical blocks. A good separation point for logical blocks could be the if conditions, but that is of course not necessary as there may be some redundant logic there too. When you model the buy side and sell side it will be easy to identify what the overlap is.

This is one the main reasons whey we needed extensive testing of this function with numerous test cases so we know that there are no regressions when we do such a significant refactor.

does this make sense? would be useful to discuss in our 1-1 tomorrow as well.

plugins/volumeFilter_test.go
```diff
@@ -155,130 +155,130 @@ func TestMakeFilterVolume(t *testing.T) {
...
-		otbQuote           *float64
-		tbbBase            *float64
-		tbbQuote           *float64
-		inputOp            *txnbuild.ManageSellOffer
-		wantOp             *txnbuild.ManageSellOffer
-		wantTbbBase        *float64
-		wantTbbQuote       *float64
+		name           string
+		mode           volumeFilterMode
+		baseCapInBase  *float64
```
nikhilsaraf, line 160: PR submission tip:
usually a good idea to do renames in a completely separate PR, even if it's only a 1-line PR. such PRs should include renames of only one type so it's a 1-second review vs something longer, and so it doesn't get held up with other changes such as the one above.

example: `sellBaseCapInBase` -> `baseCapInBase` + `sellBaseCapInQuote` -> `baseCapInQuote` can be in 1 PR, however a hypothetical rename of `sellBaseCapInBase` -> `sellBaseCapInTop` + `sellBaseCapInQuote` -> `sellBaseCapInBottom` should be in two separate PRs to make the review process much much simpler.

does that make sense?

if you prefer extracting out the renames into a separate PR that's ok with me, or if you want to leave here.

nikhilsaraf (commented):
let's discuss in our 1-1 slot tomorrow

plugins/volumeFilter_test.go
debnil: Yup, this makes a lot of sense. As discussed, I'm changing this PR to just the renames, and I'll roll the buy-side implementation into a separate PR.

plugins/volumeFilter.go
debnil: Ah, got it, ok. I'll revert these names as well.

nikhilsaraf (approved):
lgtm, thanks! 🚢

plugins/volumeFilter.go
debnil: Yup, it makes a ton of sense and is super helpful. Now that we've made the associated renames in this PR, I'll work on pulling out that logic in a follow-up :)

PR #602, opened 2020-12-04
Refactor capacity compute in volumeFilterFn

debnil:
This PR refactors the capacity computation within `volumeFilterFn`. It creates a new function, `getRemainingCapacity`, that takes the sell offer amount and the various caps, and returns whether or not the amount should be updated. That function is intended to improve readability and reduce duplicate, hand-written logic.

### DIFF
--- a/plugins/volumeFilter.go
+++ b/plugins/volumeFilter.go
@@ -158,6 +158,11 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
 }
 
 func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFilterConfig, op *txnbuild.ManageSellOffer, baseAsset hProtocol.Asset, quoteAsset hProtocol.Asset, lp limitParameters) (*txnbuild.ManageSellOffer, error) {
+	// if not in exact mode, return a nil op
+	if lp.mode != volumeFilterModeExact {
+		return nil, nil
+	}
+
 	isSell, e := utils.IsSelling(baseAsset, quoteAsset, op.Selling, op.Buying)
 	if e != nil {
 		return nil, fmt.Errorf("error when running the isSelling check for offer '%+v': %s", *op, e)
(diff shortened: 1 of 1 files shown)


plugins/volumeFilter.go
```diff
@@ -158,6 +158,11 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
 }
 
 func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFilterConfig, op *txnbuild.ManageSellOffer, baseAsset hProtocol.Asset, quoteAsset hProtocol.Asset, lp limitParameters) (*txnbuild.ManageSellOffer, error) {
+	// if not in exact mode, return a nil op
```
debnil, line 161: From the previous function, if the mode was not `Exact`, a `nil` op would be returned. I wasn't sure if that was a desired invariant of the function or a side-effect of its implementation. So I'm just flagging this early return to confirm correctness.

plugins/volumeFilter.go
```diff
@@ -228,6 +241,17 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
 	return nil, nil
 }
 
+func getRemainingCapacity(otbCap float64, tbbCap float64, limitCap float64, amount float64, price float64) float64 {
```
debnil, line 244: This function is a direct implementation of the previous capacity computations ([base](https://github.com/stellar/kelp/pull/602/files#diff-21b706ce390d2571295650103c3c1953c7d81665c8b2672c9de2208dc9541952L182-L192), [quote](https://github.com/stellar/kelp/pull/602/files#diff-21b706ce390d2571295650103c3c1953c7d81665c8b2672c9de2208dc9541952L200-L210)).

(Note that when invoked above, in the base if-statement, the price is passed as 1 - thus making this implementation generalizable in either case.)

plugins/volumeFilter.go
```diff
@@ -174,47 +179,55 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
...
+			// if the amount changed, update the amount and the log
+			if newAmountBeingSold != amountValueUnitsBeingSold {
+				newAmount = fmt.Sprintf("%.7f", newAmountBeingSold)
+				newAmountLog = fmt.Sprintf(", newAmountString = %s", newAmount)
 			}
-			log.Printf("volumeFilter:  selling (base units), price=%.8f amount=%.8f, keep = (projectedSoldInBaseUnits) %.7f <= %.7f (config.BaseAssetCapInBaseUnits): keepSellingBase = %v%s", sellPrice, amountValueUnitsBeingSold, projectedSoldInBaseUnits, *lp.baseAssetCapInBaseUnits, keepSellingBase, newAmountString)
-		} else {
-			keepSellingBase = true
+
+			// TODO DS Confirm that this is the desired approach to logging.
```
debnil, line 201: The approach to logging is my only real open question. I tried to include all the information in the existing logs, excluding `keepSellingBase`/`keepSellingQuote`, as those variables no longer exist. This required computing `projectedSoldInBaseUnits` here as well as in the new extracted function. I considered logging in the extracted function, but that pretty clearly solves it at the wrong level. Curious if you have thoughts here - should I change logging entirely?

plugins/volumeFilter.go
nikhilsaraf: This is not correct.

If the mode is `Ignore` it will not return `nil` but will return a non-nil op.
If it was to return `nil` then anyone who used the `Exact` feature would never place any orders as we would drop all of them before we placed it!

This demonstrates an incorrect understanding of what is going on in this PR. does not make sense to review the rest of this PR.

nikhilsaraf (commented):
The first line of this review is grossly incorrect and would severely regress the codebase.

Let's close this PR until the following is done:
1. add missing test cases (see below)
2. we have a flowchart / decision tree that captures the logic in the volumeFilterFn so we don't regress the codebase.

----

The following test cases are missing in `TestVolumeFilterFn` (all cases where cap > tbb + offer) which would have caught it. We should add all combinations explicitly to the set of test cases.

one example:
```
{
			name:           "11. selling, quote units sell cap, keep selling quote, ignore mode",
			mode:           volumeFilterModeIgnore,
			baseCapInBase:  nil,
			baseCapInQuote: pointy.Float64(1000.0),
			otbBase:        pointy.Float64(0.0),
			otbQuote:       pointy.Float64(0.0),
			tbbBase:        pointy.Float64(0.0),
			tbbQuote:       pointy.Float64(0.0),
			inputOp:        makeManageSellOffer("2.0", "100.0"),
			wantOp:         nil,
			wantTbbBase:    pointy.Float64(0.0),
			wantTbbQuote:   pointy.Float64(0.0),
		},
```
This should have been covered in the previous PRs, I think I added a comment about this but it was not addressed. There are still some cases that are not tackled in that in addition to what is mentioned above such as non-zero values for `tbbQuote` etc. -- there are a lot of 0-value test cases in that test which is not very valuable.

When looking at those test cases please remove the `for` loop around `action`. we will be adding explicit test cases for both actions separately in those test cases.

Not having good test cases can break the code quite significantly, let's try and have a little more attention to detail.

debnil:
Will do both. I can't find either of these requests explicitly made in comments on past PRs. Could we add guidelines to the `Contributing Guidelines` around follow-up requests - so it's clear whether the reviewer or submitter should add TODOs, file issues, etc? It was definitely not clear to me that these were action items on my plate. That seems super fixable through a clearer norm around filing follow-up issues.

1. Flowchart:
> Before we submit a PR, please upload that onto the Github issue. I had requested this previously as well.
Where was a flowchart requested? The only reference to flowcharts is in [this comment](https://github.com/stellar/kelp/pull/600#discussion_r534531934). This definitely doesn't request a flowchart; I don’t have this request in my notes from our sync yesterday; and we don't have an issue for this refactor anyway. The next step we talked about was this functional refactor. What am I missing here?

In any case, this comment was really helpful, so I drew out a flow chart for my own understanding. While there is no issue for this item, I can make one, digitize said flow chart, and put it there. The flow chart directly resulted in this early return. I'll outline my logic in a subsequent comment.

2. Tests where cap > tbb + offer
> The following test cases are missing in `TestVolumeFilterFn` (all cases where cap > tbb + offer) which would have caught it. We should add all combinations explicitly to the set of test cases. 
> This should have been covered in the previous PRs, I think I added a comment about this but it was not addressed. There are still some cases that are not tackled in that in addition to what is mentioned above such as non-zero values for `tbbQuote` etc. -- there are a lot of 0-value test cases in that test which is not very valuable.

I definitely missed adding the non-zero TBB cap cases requested [here](https://github.com/stellar/kelp/pull/585#pullrequestreview-532781672) - but I don't see which comment explicitly requests `cap > tbb + offer`? In any case, a clearer process around follow-up issues can help avoid this confusion.

plugins/volumeFilter.go
debnil: My bad on this.  Ah, I tried to redo my chart and see what I did wrong here. Definitely a strong argument for this refactor.

For fuller context, I had put that we kept selling in two cases:
1. Nil cap.
2. Projected sold exceeds cap; mode is exact; and new amount is positive and changed on the operation

Of course, this misses the obvious case:
3. Projected sold is under the cap. In that case, even if the mode is `Ignore`, then `keepSellingBase`/`keepSellingQuote` would remain true.

However, I think one point of conversation in our sync yesterday might then be incorrect. We said that the mode can be checked outside the function that computes remaining capacity. If we do this, we must double compute the projected sold amount, and we lose much of the benefit of this abstraction. I'll think more about this.

debnil:
@nikhilsaraf - did you run the test case you suggested on `master`? I just ran it, and it already fails. Is the test incorrect; is the existing implementation wrong; or is my `master` out-of-sync?

I'm pretty sure the test is incorrect - if `wantOp` would be `nil`, this refactor would be correct. But the clarification would be helpful, as the action item is to write more tests like this one. Thanks!

```
expected: (*txnbuild.ManageSellOffer)(nil)
            	            	actual  : &txnbuild.ManageSellOffer{Selling:txnbuild.NativeAsset{}, Buying:txnbuild.NativeAsset{}, Amount:"100.0", Price:"2.0", OfferID:0, SourceAccount:txnbuild.Account(nil)}
```

PR #614, opened 2020-12-09
Add volume filter tests for non-nil quote and ignore mode

debnil:
This PR adds the tests for `volumeFilterFn` when the quote cap is non-nil and the mode is `ignore`. This extends #611 to when the quote cap is non-nil, rather than when the base is. Follow-up PRs will deal with the `exact` mode.


nikhilsaraf (approved):
LGTM, awesome! 🎉

PR #615, opened 2020-12-09
Add tests for volume filter for non-nil base and exact mode

debnil:
This PR adds tests for the volume filter for when the base cap is non-nil and the mode is `Exact`. Follow up to #611 and #614. Follow-up PR will add tests for non-nil quote and exact.


plugins/volumeFilter_test.go
```diff
@@ -166,6 +166,277 @@ type volumeFilterFnTestCase struct {
...
+			tbb:          0,
+			inputPrice:   2.0,
+			inputAmount:  15.0,
+			wantPrice:    nil,
+			wantAmount:   nil,
+			wantTbbBase:  0,
+			wantTbbQuote: 0,
+		},
+		{
+			name:         "15. otb > 0 && tbb > 0; projected < cap, newAmount = 0",
```
nikhilsaraf, line 351: there may be an error with the name here, newAmount is not 0 in this test case.

I'm wondering if it was leftover from copy-pasting of a previous case, can we rename this to remove `newAmount = 0`?

plugins/volumeFilter_test.go
```diff
@@ -166,6 +166,277 @@ type volumeFilterFnTestCase struct {
...
+			cap:          0.0,
+			otb:          0,
+			tbb:          0,
+			inputPrice:   2.0,
+			inputAmount:  15.0,
+			wantPrice:    nil,
+			wantAmount:   nil,
+			wantTbbBase:  0,
+			wantTbbQuote: 0,
+		},
```
nikhilsaraf, line 349: can you add a comment on a line between the 14th and 15th case here, saying something like:
we cannot test when otb = 0 && tbb = 0 and when newAmount < 0 so we are skipping that case.

when following the cases one by one this will be helpful in realizing that it was skipped intentionally.

nikhilsaraf (commented):
added two small inline comments, almost done!

plugins/volumeFilter_test.go
debnil: Thanks for catching - just removed `newAmount = 0`.

plugins/volumeFilter_test.go
```diff
@@ -349,7 +349,7 @@ func TestVolumeFilterFn_BaseCap_Exact(t *testing.T) {
 		},
 		// it is not possible for otb = 0 && tbb = 0 and newAmount < 0, so skipping that case
 		{
-			name:         "15. otb > 0 && tbb > 0; projected > cap",
+			name:         "15. otb > 0 && tbb > 0; projected < cap",
```
nikhilsaraf, line 352: @debnil this case is the `<` case, edited to fix

nikhilsaraf (approved):
LGTM, thanks! 🎉

PR #616, opened 2020-12-10
Add volume filter tests for non-nil quote cap and exact mode

debnil:
This PR adds tests for the volume filter for non-nil quote cap and exact mode. It's a follow up to #611 and #614.

### DIFF
--- a/plugins/volumeFilter_test.go
+++ b/plugins/volumeFilter_test.go
@@ -168,6 +168,19 @@ func TestVolumeFilterFn(t *testing.T) {
 		wantTbbBase    *float64
 		wantTbbQuote   *float64
 	}{
+		// These are all the paths that can happen, assuming exactly one quote exists (a check before this function).
+		//
+		// no quote cap; base cap, projected sold < cap -> keep selling base w/ same amount
+		// no quote cap; base cap, projected sold = cap -> keep selling base w/ same amount
+		// no quote cap; base cap, projected sold > cap, not exact -> don't keep selling base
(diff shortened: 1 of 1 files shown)


plugins/volumeFilter_test.go
```diff
@@ -166,6 +166,278 @@ type volumeFilterFnTestCase struct {
...
+
+		runTestVolumeFilterFn(
+			t,
+			k.name,
+			volumeFilterModeExact,
+			queries.DailyVolumeActionSell,
+			nil,                   // base cap nil because this test is for the QuoteCap
+			pointy.Float64(k.cap), // quote
+			nil,                   // baseOTB nil because this test is for the QuoteCap
+			pointy.Float64(k.otb), // quoteOT
```
nikhilsaraf, line 430: since we're making changes on this, can we also fix the typo in this comment to say `quoteOTB`?

nikhilsaraf (commented):
minor comments, sound be merged in next iteration. good stuff! 🎉

nikhilsaraf:
@debnil looks like there's a conflict with the master branch.

I think there's no actual overlapping code but the two new tests cases were placed in the same spot so the merge logic is having a hard time. I'd suggest doing one of the following to make this easiest:

1. resolve outstanding issues, get approval on this PR, and then resubmit as a new PR with a cut-paste of this test based off the master branch
1. resubmit as a new PR with a cut-paste of this test based off the master branch without resolving any issues, then in a separate commit resolve outstanding issues so its easier to identify the fixes to these comments, get approval on this PR and merge

PR #617, opened 2020-12-10
Delete previous volume filter tests

debnil:
This PR deletes the old set of volume filter tests, now that we have rewritten them to be fully comprehensive. This should be merged after #615 and #616 are approved.


nikhilsaraf (approved):
lgtm, thank you! 🎉

PR #620, opened 2020-12-11
Add volume filter tests for non-nil quote cap and exact mode

debnil:
This PR adds the tests for the volume filter for a non-nil quote cap and exact mode. It replaces #616 due to merge conflicts.


nikhilsaraf (approved):
this looks great, thank you for resubmitting as a new PR! 🎉

Issue #604, opened 2020-12-05
Refactor volume filter function


debnil:
The `volumeFilterFn` within `plugins/volumeFilter.go` can be significantly streamlined. The below flowchart outlines its logical flow to help guide a refactor.

![Volume filter function flowchart](https://user-images.githubusercontent.com/1740974/101227686-9aa7c500-364d-11eb-940d-c09201696085.png)

debnil:
When looking at the above flowchart, we’re basically trying to answer: when should either the base or quote asset keep being sold?
1. If the relevant cap is nil. Then, we return the operation with the same amount.
2. The projected amount sold is within the cap’s value. Then, we return the operation with the same amount.
3. If the mode is exact and there’s still remaining capacity. Then, we update the amount on the operation and return it.

If one of the above conditions is not met for exactly one asset, we return no operation.

nikhilsaraf:
Flowchart is awesome, well done! 💯

debnil:
This reworked/simplified version of the above flowchart converts a buying op to a selling op and formats a base cap as a quote cap. This simplifies it down to a single workflow.

![Reworked flowchart](https://user-images.githubusercontent.com/1740974/102139058-2cfc5580-3e12-11eb-939c-fe142acc9dfc.png)

nikhilsaraf:
1. we probably want to update the op amount only in the case where there is remaining capacity (bottom right of second flowchart). In the case where there is no remaining capacity we should discard the op (i.e. don't keep selling quote).

2. I'm not sure if we can convert a base cap to a quote cap. I think the more appropriate thing to do is to "select otb, tbb, cap from base part of inputs" or from the quote part of inputs. I'm thinking that the common format we select will not have any idea of "base" or "quote" but only has an understanding of otb, tbb, cap, and offerAmount. Any conversions/transformations/selections before that should provide this invariant. What do you think?

3. we probably need logic (box) to convert sell offers back to buy offers. What is the best place to put this? based on the Tip below do we even need this?

4. we also need a box to represent where and when we will update the TBB values. should we do this before converting back to a buy op or before? (i.e. what is the unit of accounting here)

------

Tip:
- a "sell base" limit is a limit on the amount of base asset that is sold as a result of selling the base asset.
- a "sell quote" limit is a limit on the amount of base asset that is sold denominated in the quote asset, i.e. the amount of quote asset that is purchased as a result of selling the base asset.
[...]

PR #625, opened 2020-12-16
Refactor volume filter function (part of #604)

debnil:
This PR refactors the volume filter function - it streamlines duplicate functionality and also handles buy ops. Part of #604.


plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
+		return nil, fmt.Errorf("could not compare offer and filter: %s", e)
 	}
 
-	sellPrice, e := strconv.ParseFloat(op.Price, 64)
+	if !isFilterApplicable {
+		// ignore filter so return op directly
+		return op, nil
+	}
+
+	offerPrice, e := strconv.ParseFloat(op.Price, 64)
```
nikhilsaraf, line 188: not sure if op.Price is inverted in the case of a buy because of the complexities of how manage sell offer works. It's complicated and not something we need to get right in this PR.

but this is something we should look at when working on the tests for the buy side. Can you add a TODO DS comment here so we remember to get this right in the next PR?

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
-				}
-			}
-			log.Printf("volumeFilter: selling (quote units), price=%.8f amount=%.8f, keep = (projectedSoldInQuoteUnits) %.7f <= %.7f (config.BaseAssetCapInQuoteUnits): keepSellingQuote = %v%s", sellPrice, amountValueUnitsBeingSold, projectedSoldInQuoteUnits, *lp.baseAssetCapInQuoteUnits, keepSellingQuote, newAmountString)
-		} else {
-			keepSellingQuote = true
-		}
+	// if projected is under the cap, update the tbb and return the original op
+	projected := otb + tbb + offerAmount*capPrice
+	if projected <= cap {
+		*dailyTBBAccumulator.BaseAssetCapInBaseUnits += offerAmount
```
nikhilsaraf, line 214: can you extract these two lines in it's own function. we repeat it below so we can just reuse this function. reduces chances of errors so we don't update one and not the other by mistake in the future

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
-	// we don't want to keep it so return the dropped command
-	return nil, nil
+	// if exact mode and with remaining capacity, update the op amount and return the op
+	// else, return nil
+	remainingCapacity := (cap - otb - tbb) / capPrice
+	if remainingCapacity <= 0 {
+		return nil, nil
+	}
+
+	offerAmount = remainingCapacity
```
nikhilsaraf, line 230: is it correct to call this `offerBaseAmount` because we divide by `capPrice`? if so then let's use that name to be more explicit since we have the context on it now but not when we will revisit the code next in the future

nikhilsaraf: PS: we should try to avoid reusing variables where possible. Using new variable names as a general practice makes the code more readable since developers don't need to remember what this variable was assigned to previously and if we ever reuse it again.

A good name for this to keep it separate from `offerAmount` would be either `newOfferAmount` or `newOfferBaseAmount`.

nikhilsaraf: PPS: that being said, we probably don't need a separate variable for `offerAmount` or `offerAmountBase` since it's the same thing as `remainingCapacity` which we can probably rename to `remainingCapacityBase`. alternatively we can rename `remainingCapacity` to `offerAmount` or `offerAmountBase` accordingly.
i.e. we don't do anything to `offerAmount` here so this is effectively just a rename which may not be required

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
+	op.Amount = fmt.Sprintf("%.7f", offerAmount)
+	*dailyTBBAccumulator.BaseAssetCapInBaseUnits += offerAmount
```
nikhilsaraf, line 232: we could use the common function that we will extract above.

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
-		// TODO buying side - we need to implement this to support buy side filters; extract common logic from the above sell side case
+	if lp.mode != volumeFilterModeExact {
+		return nil, nil
 	}
 
-	// we don't want to keep it so return the dropped command
-	return nil, nil
+	// if exact mode and with remaining capacity, update the op amount and return the op
+	// else, return nil
+	remainingCapacity := (cap - otb - tbb) / capPrice
```
nikhilsaraf, line 225: here too will need to tweak for buy side if it is different because of the complexities of how manage sell offer works. can add as a TODO if you want (like suggested above)

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
+	}
+
+	offerAmount = remainingCapacity
+	op.Amount = fmt.Sprintf("%.7f", offerAmount)
+	*dailyTBBAccumulator.BaseAssetCapInBaseUnits += offerAmount
+	*dailyTBBAccumulator.BaseAssetCapInQuoteUnits += offerAmount * offerPrice
+	return op, nil
+}
+
+func offerSameTypeAsFilter(filter *VolumeFilterConfig, op *txnbuild.ManageSellOffer, baseAsset hProtocol.Asset, quoteAsset hProtocol.Asset) (bool, error) {
```
nikhilsaraf, line 237: 👍

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
+	opIsSelling, e := utils.IsSelling(baseAsset, quoteAsset, op.Selling, op.Buying)
+	if e != nil {
+		return false, fmt.Errorf("error when running the isSelling check for offer '%+v': %s", *op, e)
+	}
+	filterIsSelling := filter.action == queries.DailyVolumeActionSell
```
nikhilsaraf, line 242: this logic should live on the type. can you please update `dailyVolumeByDate.go` (~line 31) with a function on `DailyVolumeAction` that returns `isSell() bool` and `isBuy() bool`?

we can use is here by saying (explicitly using the `isSame` variable name makes it more readable too):
```
isSame := opIsSelling == filter.action.IsSell()
return isSame, nil
```

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
+	return opIsSelling == filterIsSelling, nil
+}
+
+func extractFilterInputs(dailyOTB *VolumeFilterConfig, dailyTBB *VolumeFilterConfig, lp limitParameters) (float64, float64, float64, error) {
```
nikhilsaraf, line 246: can you add a comment on this function. ideally we should use a better function name instead of `extractFilterInputs`

nikhilsaraf, line 246: whenever we return values of the same type we should comment what is being returned. the comment should be of the form /* name */ and should come *after* the type -- that's just how comments are best formatted in such a situation. you can leave out comments for non-repeated types such as error in this case

plugins/volumeFilter.go
```diff
@@ -175,74 +175,95 @@ func (f *volumeFilter) Apply(ops []txnbuild.Operation, sellingOffers []hProtocol
...
+
+	if lp.baseAssetCapInQuoteUnits != nil {
+		return *dailyOTB.BaseAssetCapInQuoteUnits, *dailyTBB.BaseAssetCapInQuoteUnits, *lp.baseAssetCapInQuoteUnits, nil
+	}
+
+	// should never reach this code - means that the configs were not validated properly
+	return -1, -1, -1, fmt.Errorf("found two nil filters")
+}
+
+func updateOpBasedOnCap(op *txnbuild.ManageSellOffer, otb float64, tbb float64, cap float64, price float64) (*txnbuild.ManageSellOffer, float64) {
```
nikhilsaraf, line 259: don't think this function `updateOpBasedOnCap` is used since this is handled in the main filter function. can remove

nikhilsaraf (commented):
this is a very good refactor, and great to see all the tests are passing!

some comments inline, should be ready to merge in next iteration. thank you!

plugins/volumeFilter.go
debnil: Yeah, this all makes sense. I think you're right that we don't need to assign this to `offerAmount`. I'm just going to keep it as `remainingCapacity`.

plugins/volumeFilter.go
```diff
@@ -222,15 +222,14 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
 
 	// if exact mode and with remaining capacity, update the op amount and return the op
 	// else, return nil
-	remainingCapacity := (cap - otb - tbb) / capPrice
-	if remainingCapacity <= 0 {
+	// TODO DS Determine whether this calculation works for a buy offer.
+	offerAmount = (cap - otb - tbb) / capPrice
```
nikhilsaraf, line 226: can we call this `newOfferAmount` as a new variable instead of reusing the `offerAmount` variable?

nikhilsaraf (commented):
one var rename request, rest all looks great!

nikhilsaraf (approved):
LGTM, thank you ! 🎉

PR #626, opened 2020-12-17
Test buy-side for volume filter with base cap and exact mode

debnil:
This PR integrates tests for the buy-side for the volume filter, with the base cap and exact mode. Since we are keeping the test cases the same as the sell-side, this PR extends the existing function to also test base. It creates a buy operation with a different base asset; changes the asset used in `runTestVolumeFilterFn`; and loops over both actions in the test function.

To make the existing test cases all work, this also tweaks the `makeBuyOpSellAmt` and the logic around computing buy offers.

### DIFF
--- a/plugins/volumeFilter.go
+++ b/plugins/volumeFilter.go
@@ -191,16 +191,24 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
 		return nil, fmt.Errorf("could not convert price (%s) to float: %s", op.Price, e)
 	}
 
+	offerAmount, e := strconv.ParseFloat(op.Amount, 64)
+	if e != nil {
+		return nil, fmt.Errorf("could not convert amount (%s) to float: %s", op.Amount, e)
+	}
+
+	// A "buy" op has amount = sellAmount * sellPrice, and price = 1/sellPrice
+	// So, we adjust the offer variables by "undoing" those adjustments
+	// We can then use the same computations as sell orders on buy orders
+	if dailyOTB.action.IsBuy() {
+		offerAmount = offerAmount * offerPrice
+		offerPrice = 1 / offerPrice
+	}
+
 	// capPrice is used when computing amounts to sell or buy
 	// it's the offer price when capping on quote, and 1.0 when capping on base
 	capPrice := offerPrice
 	if lp.baseAssetCapInBaseUnits != nil {
 		capPrice = 1.0
-	}
-
-	offerAmount, e := strconv.ParseFloat(op.Amount, 64)
-	if e != nil {
-		return nil, fmt.Errorf("could not convert amount (%s) to float: %s", op.Amount, e)
 	}
 
 	// extracts from base or quote side, depending on filter
@@ -228,8 +236,15 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
 		return nil, nil
 	}
 
-	op.Amount = fmt.Sprintf("%.7f", newOfferAmount)
 	dailyTBBAccumulator = updateTBB(dailyTBBAccumulator, newOfferAmount, offerPrice)
+
+	// if we have a buy operation, we want to make sure buy ops have the same relationship between price and amount
+	// to do this, we apply the same amount adjustment as `makeBuyOpAmtPrice`
+	newOpAmount := newOfferAmount
+	if dailyOTB.action.IsBuy() {
+		newOpAmount = newOpAmount * offerPrice
+	}
+	op.Amount = fmt.Sprintf("%.7f", newOpAmount)
 	return op, nil
 }
 
--- a/plugins/volumeFilter.go
+++ b/plugins/volumeFilter.go
@@ -240,6 +240,13 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
 
 	// if we have a buy operation, we want to make sure buy ops have the same relationship between price and amount
 	// to do this, we apply the same amount adjustment as `makeBuyOpAmtPrice`
+	// The following conversion is done above on input:
+	// sellOfferAmount = buyOfferAmount * buyOfferPrice
+	// sellOfferPrice = 1 / buyOfferPrice
+	//
+	// Therefore we need to undo it using the following:
+	// newOpAmount = newOpAmount * sellOfferPrice
+	// newOpAmount => newOpAmount * 1 / buyOfferPrice
 	newOpAmount := newOfferAmount
 	if dailyOTB.action.IsBuy() {
 		newOpAmount = newOpAmount * offerPrice
--- a/plugins/volumeFilter_test.go
+++ b/plugins/volumeFilter_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var testBaseAsset = txnbuild.NativeAsset{}
 var testQuoteAsset txnbuild.CreditAsset = txnbuild.CreditAsset{Code: "QUOTE", Issuer: "GBGQAGAMK6W6FH6AGGZ2BI2MY5TA5VJEHU2DQRFXACMAZHNRD3SXEV6Z"}
 
 func makeWantVolumeFilter(config *VolumeFilterConfig, marketIDs []string, accountIDs []string, action queries.DailyVolumeAction) *volumeFilter {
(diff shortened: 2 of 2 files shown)


nikhilsaraf (commented):
Adding comments as a bit more general here for faster review turnaround on my end:
1. Can we avoid the for loops with the action variable? Better to be more explicit. We can have two invocations of runVolumeFilterTest (one for sell and the other for buy). This would remove the makeActionOpAmtPrice function as well and would increase readability of the test. What do you think?
2. I was thinking that we would need a wantSellAmt/Price and a wantBuyAmt/Price variables as inputs to the test case instead of a common wantAmt/Price mostly because I wasn’t expecting the numbers to work out to be exactly the same. Will think more about this when I look at it again tomorrow. Is this always the case across all types of test combinations? Didn’t want to hold up the review for this.
3. We should not use an invalid combination of assets such as native, native for selling. Let’s use native as the base for selling and a new asset (QUOTE instead of COUPON) for the quote asset. These can be declared at the top of the test file. When we make the buying op we would use QUOTE for the base field (as you’ve done) and the comment should say that a buy op is buying the base and selling the quote so we create a sell op here to sell quote and buy base.

Seems pretty straightforward otherwise. Will get back to you about # 2, but no need to hold up the reversion for this (if you have thoughts on why it’s the same please let me know so I can use that as context when looking at it tomorrow).

debnil:
Awesome, all of these are pretty straightforward changes. I can make 1 and 3 quite quickly. I'm not certain about 2 either, more below.

1. Totally on board with this. Two invocations per test case makes sense to me - will make this change.
2. Yes, always the exact same across cases. But the reason for this is that we treat the price the same in both buy and sell. If we do that, there's no reason any values would change. I'm not certain whether this is the correct approach to price. I think the key question is - how does the function that generates `ManageSellOffer` ops before `volumeFilterFn` view the notion of price? As long as that function and this one align, we are fine.
3. Cool, this makes sense to me as well.

plugins/volumeFilter_test.go
```diff
@@ -1111,7 +1139,12 @@ func runTestVolumeFilterFn(
 			mode:                     mode,
 		}
 
-		actual, e := volumeFilterFn(dailyOTB, dailyTBBAccumulator, inputOp, utils.NativeAsset, utils.NativeAsset, lp)
+		// For our tests, the selling asset is QUOTE, and the buying asset is XLM
+		// To identify a sell op, we check if the base is the selling asset and the quote is buying
```
nikhilsaraf, line 1143: This part confuses me a little.

1. The `volumeFilterFn` isn't aware of whether we are sell mode or buy mode but is more concerned with base and quote asset, so naming the parameters to the function as baseAsset and quoteAsset would be a little less confusing.
1. I was expecting the base asset to always be the native asset and the quote asset to be the quote asset. I think these are passed in reverse over here, but what makes this particularly confusing is that the test is passing this way. I think the comments below will help fix this

plugins/volumeFilter_test.go
```diff
@@ -1127,6 +1160,16 @@ func runTestVolumeFilterFn(
 func makeSellOpAmtPrice(amount float64, price float64) *txnbuild.ManageSellOffer {
 	return &txnbuild.ManageSellOffer{
 		Buying:  txnbuild.NativeAsset{},
+		Selling: testQuoteAsset,
```
nikhilsaraf, line 1163: we should be selling the base asset here -- base asset should be native. apologies if I indicated otherwise previously

plugins/volumeFilter_test.go
```diff
@@ -1127,6 +1160,16 @@ func runTestVolumeFilterFn(
...
+		Amount:  fmt.Sprintf("%.7f", amount),
+		Price:   fmt.Sprintf("%.7f", price),
+	}
+}
+
+func makeBuyOpAmtPrice(amount float64, price float64) *txnbuild.ManageSellOffer {
+	// Since a buy op buys the base and sells the quote, we reverse the assets and create a sell op.
+	return &txnbuild.ManageSellOffer{
+		Buying:  testQuoteAsset,
```
nikhilsaraf, line 1172: we should be buying the base asset here and selling the quote -- base asset should be native.
the comment you've added here is helpful but we can remove the part after the comma.

context: the fact that it's a ManageSellOffer doesn't matter because we explicitly specify the buying and selling assets. It makes more sense when thinking about it in terms of base and quote.

the way kelp works (and feel free to add this line as a comment in makeBuyOpAmtPrice function if you find this context to be more helpful than confusing) is that all actions are performed in the context of the base asset. So a sell op will SELL the BASE asset. A buy up will BUY the BASE asset. It helps to name the assets as base and quote so it's easier to think in terms of base vs. quote instead of specific assets.

on the top can you add a variable called testBaseAsset setting that to the native asset to simplify things?

nikhilsaraf (commented):
@debnil I'm not yet sure if the numbers will work out correctly (# 2) in my comment above, but I've added some comments around the assets we use. I feel this may affect the numbers a little.

if it does then we should split up the `wantPrice` and `wantAmount` into:
- `wantSellPrice`
- `wantSellAmount`
- `wantBuyPrice`
- `wantBuyAmount`

plugins/volumeFilter_test.go
debnil: Yup, the comments below fixed this. Makes sense, since it just switches the buying and assets in the operations - which means that switching the parameters here also ensures correctness.

plugins/volumeFilter_test.go
debnil: This makes sense! Added a comment with some of this info, and put a `testBaseAsset` variable at the top.

plugins/volumeFilter_test.go
```diff
@@ -1126,8 +1157,19 @@ func runTestVolumeFilterFn(
...
-		Buying:  txnbuild.NativeAsset{},
-		Selling: txnbuild.NativeAsset{},
+		Buying:  testQuoteAsset,
+		Selling: testBaseAsset,
+		Amount:  fmt.Sprintf("%.7f", amount),
+		Price:   fmt.Sprintf("%.7f", price),
+	}
+}
+
+func makeBuyOpAmtPrice(amount float64, price float64) *txnbuild.ManageSellOffer {
```
nikhilsaraf, line 1167: I'm a little confused about this function, rest looks perfect (buy offers are tricky!)

here we are selling `amount` units of of the selling asset, i.e. of the quote asset. But what we want is for the `amount` to represent the _units of the base asset_. To achieve this we will need to do some math in this function to figure out what value to set as the amount field on mso (which represents quote units) so that the base units is equal to the `amount` input.

I ran this using the buysell strategy with the amount set to 1000.0 units (of base on either side), with the buy op being displayed first:

```
2020/12/23 02:35:59 submitting the following ops
2020/12/23 02:35:59 &{Selling:{Code:COUPON Issuer:GBMMZMK2DC4FFP4CAI6KCVNCQ7WLO5A7DQU7EC7WGHRDQBZB763X4OQI} Buying:{} Amount:163.4735274 Price:6.1171984 OfferID:0 SourceAccount:0xc00047e5e0}
2020/12/23 02:35:59 &{Selling:{} Buying:{Code:COUPON Issuer:GBMMZMK2DC4FFP4CAI6KCVNCQ7WLO5A7DQU7EC7WGHRDQBZB763X4OQI} Amount:1000.0000000 Price:0.1638006 OfferID:0 SourceAccount:0xc00047e6a0}
```

above, we can see that the selling operation (second op) has amount = 1000.0 and price = 0.16 which is correct
the buying op (first op) has amount set to 163.xx and price set to 6.xx
we need to replicate this.

To do so, we need to:
- set our Amount field on the manageSellOffer here equal to `amount * price` (so we get 1000 * 0.16 ~= 163.xx like displayed above)
- set the Price field to `1/price` (so we get 1/0.16 ~= 6.xx like displayed above)

Going by the [definition of ManageSellOffer in the docs][definition-mso] price represents 1 unit of selling in terms of buying, or 1 unit of quote in terms of base here.
In our test, at a price of XLM/QUOTE of 0.16xx we want to buy 1000 units of base (XLM) and sell 163.xx units of quote (QUOTE).
This translates to an MSO with:
- amount = 163.xx (units of selling asset being sold)
- price = 1000/163 => 6.xx (1 unit of selling in terms of buying)
This matches what we got in our calculations above.

So to summarize, we should set the `Amount` field on MSO equal to `amount * price` and Price field to `1/price`

does this make sense?

[definition-mso]: https://www.stellar.org/developers/guides/concepts/list-of-operations.html#manage-sell-offer

nikhilsaraf: PS: please can you add the commented code sample above until the end of the first set of bullet points into the code as a comment. This will make sure we save this thinking for future reference next to the code itself. thanks!

nikhilsaraf (commented):
Thanks for making the changes. added a comment about the make buy op fn which I think nails down what we want it to do. I don't think this should affect the want numbers and more importantly I think we now have the operations correct with this change.

plugins/volumeFilter_test.go
debnil: Nice, this makes perfect sense! As discussed, this breaks a bunch of buy-side tests. I think this is because the test cases are now in a weird boundary condition, where they meet the specified condition for the sell side but not the buy side. I'm not sure if this requires tweaking the test case or making  `wantBuy`/`wantSell`  values for Amount and Price.

plugins/volumeFilter.go
```diff
@@ -191,18 +191,26 @@ func volumeFilterFn(dailyOTB *VolumeFilterConfig, dailyTBBAccumulator *VolumeFil
...
 
+	offerAmount, e := strconv.ParseFloat(op.Amount, 64)
+	if e != nil {
+		return nil, fmt.Errorf("could not convert amount (%s) to float: %s", op.Amount, e)
+	}
+
+	// A "buy" op has amount = sellAmount * sellPrice, and price = 1/sellPrice
+	// So, we adjust the offer variables by "undoing" those adjustments
+	// We can then use the same computations as sell orders on buy orders
+	if dailyOTB.action.IsBuy() {
```
nikhilsaraf, line 202: this is the best example of when we want to reuse variables

specifically, we really do want to overwrite the previous variable since those are incorrect, and the comment explains it well 👍 💯

nikhilsaraf (approved):
looks great, thank you!! 💯

PR #627, opened 2020-12-18
Add buy tests for daily volume queries

debnil:
This PR adds buy-side tests for the daily volume queries. It patterns the existing test cases, and adds test data such that these patterned cases yield meaningful results.


nikhilsaraf (approved):
this looks great, thank you @debnil ! 🎉

PR #629, opened 2020-12-23
Add all buy-side tests for volume filter

debnil:
This PR adds all remaining buy-side tests for the volume filter function (`volumeFilterFn`). It extends #626 to the remaining combinations of filter mode and action.

### DIFF
--- a/plugins/volumeFilter_test.go
+++ b/plugins/volumeFilter_test.go
@@ -622,17 +622,17 @@ func TestVolumeFilterFn_BaseCap_Ignore(t *testing.T) {
 	}
 	for _, k := range testCases {
 		// convert to common format accepted by runTestVolumeFilterFn
-		// doing this explicitly here is easier to read rather than if we were to add "logic" to convert it to a standard format
-		inputOp := makeSellOpAmtPrice(k.inputAmount, k.inputPrice)
-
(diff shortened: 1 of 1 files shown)


plugins/volumeFilter_test.go
```diff
@@ -804,27 +829,52 @@ func TestVolumeFilterFn_QuoteCap_Ignore(t *testing.T) {
...
 		runTestVolumeFilterFn(
 			t,
-			k.name,
+			sellName,
 			volumeFilterModeIgnore,
 			queries.DailyVolumeActionSell,
 			nil,                   // base cap nil because this test is for the QuoteCap
 			pointy.Float64(k.cap), // quote cap
-			nil,                   // baseOTB nil because this test is for the BaseCap
+			nil,                   // baseOTB nil because this test is for the QuoteCap
```
nikhilsaraf, line 847: looks like this comment was fixed to say QuoteCap instead of BaseCap. great catch!

It's sometimes easer to review these minor changes and not skip over them if they are highlighted in the PR with a comment, esp. since it breaks from the general pattern of the rest of the PR

nikhilsaraf (approved):
LGTM! thank you 🎉

PR #630, opened 2020-12-23
Implement buytwap strategy (closes #522)

### LINKED ISSUE #522: [8] buyTwap strategy

debnil:
This PR implements the buytwap strategy and closes #522. It builds on recent work on implementing buy-side infrastructure across the codebase, including the volume filter, to implement this strategy.


nikhilsaraf (commented):
looks good on review, but will pull the branch and run myself to check for correctness.

nikhilsaraf:
I ran the buysell strategy first to set up some offers and then ran the buytwap strategy.

I think the offers that are being produced may not be correct:
```
levels returned (side = buy ): [{0.1416530 3077.0930066}]
...
...
buy  | modify | old level=1 | new level = 1 | triggers=[price amount] | targetPriceQuote=7.05950456 | targetAmtBase=3077.09300661 | curPriceQuote=0.14165435 | lowPriceQuote=7.05245211 | highPriceQuote=7.06657114 | curAmtBase=1000.00000007 | minAmtBase=153197.06053572 | maxAmtBase=153503.76135761
```

levels returned shows price as `0.1416530` and amount as `3077.0930066` which is correct based on the rest of the logs

in the second line, `targetPriceQuote=7.05950456` when it should have been `1/7.05950456 = 0.14xx` (note that in most cases we log price in units of quote, i.e. we should expect 0.14xx in this case. So if we're seeing 7.xx then it means there's an error somewhere). in this line it logs the old price `curPriceQuote=0.14165435` which makes it clear that we're inverting the price somewhere.

running it again showed that the price was in fact set incorrectly since it produced offers for the inverted price:
```
offer | buy  | level=1 | curPriceQuote=7.05950456 | curAmtBase=3077.09300661 | pruning=false
```

see screenshot on exchange orderbook (top bid):
Screen Shot 2020-12-24 at 12 23 57 AM

not sure exactly where the error is since the code changes we made make sense. Can you look into this?

you may need to insert some printf statements against the amount and price to see where they change.
should mostly only be related to any new code added somewhere for buy twap.
the best starting point to debug this is to log around the volumeFilter to ensure that we're not modifying the offer around that (input price should be same as output price when not nil)

nikhilsaraf (approved):

PR #667, opened 2021-03-11
Feature/automate install script

BenRacicot:
This is the installation script for MacOS written by myself and Nikhil to automate some or all of Kelp's initial setup.


README.md
```diff
@@ -131,7 +133,9 @@ To compile Kelp from source:
     * `./scripts/build.sh`
 8. Confirm one new binary file exists with version information. 
     * `./bin/kelp version`
-9. Set up CCXT to use an expanded set of priceFeeds and orderbooks (see the [Using CCXT](#using-ccxt) section for details)
+9. Run the GUI
```
nikhilsaraf, line 136: let's swap step 9 and 10. we should run ccxt-rest before we run the GUI

README.md
```diff
@@ -111,7 +111,9 @@ To run the bot in simulation mode, try this command:
 
 _Note for Windows Users: You should use a [Bash Shell][bash] to follow the steps below. This will give you a UNIX environment in which to run your commands and will enable the `./scripts/build.sh` bash script to work correctly._
 
-To compile Kelp from source:
+_Note for MacOS Users: Running [install-macos.sh][install-macos-script] should automate the steps below. (including cloning the Kelp repo to the correct Go folder) Remember, manual installation of PostgreSQL and Docker are additionally required._
+
+## Manual Installation Steps
```
nikhilsaraf, line 116: can you add a section called "Automated Install for MacOS" above this section on "Manual Installation Steps"?
Then you can include the text you have in the note above as part of that section. That will give more importance to your automated script.

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
+#!/bin/bash
+# shellcheck disable=SC2016
+set -e
+
+# Set this machine's dependencies status(es) with the OK flag
```
nikhilsaraf, line 5: can you add to this comment that this variable is what we use as the return value from functions. we do so by updating the value of the OK variable before we return from the function and the calling code can read the OK variable.

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    VERSION="1.15.7"
+
+    [ -z "$GOROOT" ] && GOROOT="$HOME/go/go-install-do-not-delete"
+    [ -z "$GOPATH" ] && GOPATH="$HOME/go"
+
+    OS="$(uname -s)"
+    ARCH="$(uname -m)"
+
+    case $OS in
+        "Linux")
```
nikhilsaraf, line 64: will this run on Linux too?

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+            esac
+            PLATFORM="linux-$ARCH"
+        ;;
+        "Darwin")
+            PLATFORM="darwin-amd64"
+        ;;
+    esac
+
+    if [ -z "$PLATFORM" ]; then
+        echo "Your operating system is not supported by the script."
```
nikhilsaraf, line 90: can we add the name of the OS that is unsupported?

`echo "Your operating system '$OS' is not supported by the script."`

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+        exit 1
+    fi
+
+    if [ '/bin/bash' == `which bash` ]; then 
+        shell_profile="$HOME/.bashrc"
+        echo "bash is installed"
+    else
+        echo "Kelp requires bash"
```
nikhilsaraf, line 98: we should exit from there if the user does not have bash. a simple exit 2 should work (keeping in mind that we're using different exit codes for different errors)

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    fi
+
+    {
```
nikhilsaraf, line 101: we can check if GOROOT is defined in the bashrc file before we do this -- so it doesn't repeat it in the bashrc file.

we can achieve this by doing a grep for "export GOROOT" on the $shell_profile and seeing if it retuned any result. you may need to do something like this to check that:

``if [ `grep "export GOROOT" $shell_profile | wc -l` -eq 0 ]; # if equals 0 then we should append``

we can reuse the condition check for `-d "$GOROOT"` that you already have below this append statement

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    fi
+
+    echo "Extracting File..."
+
+    # sudo chown -R $USER: $HOME # https://github.com/golang/go/issues/27187
+    mkdir -p "$GOROOT"
+
+    tar -C "$GOROOT" --strip-components=1 -xzf "$TEMP_DIRECTORY/go.tar.gz"
+
+    # GOROOT=${GOROOT}
```
nikhilsaraf, line 135: if we dont need this commented out code then we can remove it

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    # PATH=$GOROOT/bin:$PATH
+    # GOPATH=$GOPATH
+    # PATH=$GOPATH:$PATH
+
+    mkdir -p "${GOPATH}/"{src,pkg,bin}
```
nikhilsaraf, line 140: this is pretty cool, didn't know you could do this!! 💯

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    echo -e "\nGo $VERSION was installed into $GOROOT.\nMake sure to relogin into your shell or run:"
+    echo -e "\n\ restart script to update your environment variables."
```
nikhilsaraf, line 142: do we need them to restart the script to update the environment vars?

we can just do `source ~/.bashrc` to execute the `.bashrc` file instead

README.md
```diff
@@ -111,7 +111,9 @@ To run the bot in simulation mode, try this command:
 
 _Note for Windows Users: You should use a [Bash Shell][bash] to follow the steps below. This will give you a UNIX environment in which to run your commands and will enable the `./scripts/build.sh` bash script to work correctly._
 
-To compile Kelp from source:
+_Note for MacOS Users: Running [install-macos.sh][install-macos-script] should automate the steps below. (including cloning the Kelp repo to the correct Go folder) Remember, manual installation of PostgreSQL and Docker are additionally required._
```
nikhilsaraf, line 114: just noticed that your script does cloning of the repo too -- I remember we had discussed this last time and how we can take advantage of that. that's super cool.

Can you add to your note here that for MacOS you don't need to clone the directory but can download the automated install script from here and that it will also clone the GitHub repo in the correct location -- and link to the script they should run.

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    cd $GOPATH/src/github.com/stellar/kelp
+}
+
+# After Golang install Glide
+function installGlide() {
+    echo "Installing Glide."
+    if curl --version; then
+        curl https://glide.sh/get | sh
+    elif wget --version; then
+        wget https://glide.sh/get | sh
```
nikhilsaraf, line 170: this won't work because `glide.sh` doesn't work anymore. we will need to find an alternative way to install glide, maybe using homebrew?

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+
+# After Glide install Astilectron
+function installAstilectron() {
+    go get -u github.com/asticode/go-astilectron-bundler/... 
+    go install github.com/asticode/go-astilectron-bundler/astilectron-bundler
+}
+
+function postGoInstall() {
+    cloneIntoDir
+    isGlide
```
nikhilsaraf, line 187: do we need to invoke `installGlide` here if isGlide is false?

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+    else
+        echo "Goland is not installed"
+    fi
+}
+
+
+#**********************
+# Execute the functions
+#**********************
+cd
```
nikhilsaraf, line 232: can you add a comment here why we are doing this `cd` -- IIRC it was to get to a deterministic starting working directory

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+
+isNode
+if [ $OK ]; then
```
nikhilsaraf, line 235: maybe we can call `checkDeps` here. so it executes all of these checks and we don't need this duplicate code.
then we can reuse the $OK variable and if it is false we can simply to an exit 1.

Example:
```
#checkDeps will set the OK variable to false if at least 1 dependency is missing
checkDeps
if [ $OK ]; then
    # do nothing
else
    exit 1
fi
``` 

Since we don't want to exit in the case where Golang does not exist, we can inline the go check logic from the `checkDeps` function into `postGoInstall()` right after we call `checkDeps` (line 190)

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+        echo "`git version` is installed"
+    else
+        echo "Git is not installed"
+    fi
+
+    isGo
+    if $OK; then
+        echo "`go version` is installed"
+    else
+        echo "Goland is not installed"
```
nikhilsaraf, line 224: this should say Golang instead of Goland -- Goland is a Go-specific IDE so this can be confusing

scripts/install-macos.sh
```diff
@@ -0,0 +1,268 @@
...
+if [ $OK ]; then
+    echo "Git is installed on your machine."
+else
+    echo "Git is not installed on your machine!"
+    exit 1
+fi
+
+isGo
+if $OK; then
+    echo "true for some reason"
```
nikhilsaraf, line 260: this echo statement seems confusing. we don't need to echo anything here since we echo the go version below.

nikhilsaraf (commented):
Added some comments. We should be good to merge after these changes. This is going to be amazing!

nikhilsaraf:
@BenRacicot  based on keybase conversation we can merge this as-is and I can make the suggested changes after this PR is merged. Can you please sign the CLA so we can accept this commit? thanks for the submission! :)
https://forms.gle/9FBgjDnNYv1abnKD7

PR #717, opened 2021-07-17
feat/add ttlMap + binanceExchangeWs + binanceExchangeWs.GetTickerPrice

tibrn:
# Guidelines for submitting code to Kelp

## Before Submitting a PR

1. Problem Identification
    1. ensure you understand the problem / task.
    1. ask initial questions to understand the task described in the Github issue, which may not be clearly defined.
    1. you can reach the team on keybase or by commenting on the Github issue.
    1. this is not the point to discuss concrete solutions (concrete design or implementation), but only what the problem is and the desired outcome.
    1. ask any questions related to scope of work, breakdown of tasks into subtasks and PRs.
    1. typically this should be covered in the Github issue. If it is not, then add to the issue after discussion with code owner.
    1. author needs to be comfortable with breakdown and scope of tasks. Scope should be small. Think about the overhead on the code reviewer when breaking down the tasks.

1. Research Solution
    1. conduct research on the issue at hand. Identify point of code change and any “fallout” from the change (tip: search the codebase for consumers of the function / object, and their consumers etc., until you get to a part of the code that you are well-versed with).
    1. author needs to become the “expert” on the part of the code being touched before submitting a code change.
    1. If it will take you 2 more days to figure something out by reading the code or experimenting with the codebase then do it.
        1. This is not wasted effort.
        1. It is likely that the code reviewer would use a similar process to figure it out, but this is your job and not the code reviewer’s job.
    1. If the process of becoming an “expert” on a specific area would take you more than 2 days then discuss this with the code reviewer if it is worth it, and proceed accordingly.
    1. Think about what kind of tests you would want. Are you going with a black-box testing approach, or a white-box testing approach? Why?
1. Technical Design Discussion
    1. PRs are not the best medium to conduct design discussions. The team is readily accessible and available so 1-1 meetings can be very effective.
    1. This is the best avenue to discuss your ideas for how to approach the problem.
    1. Prepare any relevant technical documentation for the design discussion and post on the Github issue and ping code owner.
        1. This can be in the form of a google document, flowchart diagram, or in rare cases a code snippet in a Draft PR.
        1. The Draft PR should not be reviewed by the code reviewer as part of the code change but can be used as a space to get feedback
        1. Resulting comments on a Draft PR can capture any guidance from the technical discussion.
        1. All Draft PRs should be discarded and never converted to a non-draft PR to avoid the overhead of comments lingering from the design discussion using the Draft PR.
    1. Schedule technical design discussion to go over the proposed solution.
    1. Include a list of tests you need to add, remove, modify. Consider all possibilities (tip: data related tests would require many test cases).
1. Implementation
    1. implement feature or bug fix. This should be the first line of code that you write for the PR.
    1. Avoid for loops to “autogenerate” test cases, unless absolutely necessary.
1. Pull Request
    1. The first version of the PR should not have any “surprises” compared to what was discussed in the technical design.
    1. See [Handling of PRs](#handling-of-prs) section below for guide on how to get a PR to merged status.
    1. See [detailed guide on submitting code-reviewer-friendly PRs](https://mtlynch.io/code-review-love/). This guide will save a lot of time for both author and reviewer and will help speed up the turnaround time on PRs.

## Handling of PRs

Typically, this is what is expected from the author of a Pull Request:

1. Each PR should correspond to the *smallest independent logical unit* of change.
    1. each PR should be prefixed with "bugfix", "feature", or "refactor".
    1. refactors should never be mixed in with logic changes or bug fixes and vice-versa.
    1. refactors that change variable names across more than 1 function should be a separate PR altogether.
1. Author should add inline commentary on the code changes to "guide" the reviewer on how to read the code change, highlighting key changes made.
    1. this should typically be 2-3 places, as anything bigger is too big of a PR and should be split up.
1. We should aim for the following benchmarks in terms of PR efficiency:
    1. In 60% of the cases the PR should be approved without any changes requested from the reviewer.
    1. In 25% of the cases the reviewer should add 2-3 comments that need to be "fixed up". This is most likely a code style suggestion.
    1. In 10% of the cases the reviewer should request a structural change to the code. These should not be "new ideas".
    1. In <=5% of the cases the PR may be discarded because we have learned something new in the process of implementation.
1. Author makes requested changes.
    1. no "new ideas" to be introduced at this stage.
    1. no structural changes to be introduced beyond what is requested.
    1. focus should be on minimizing "changes" beyond what is requested by reviewer.
    1. person who created a comment / raised a question on the PR is responsible to mark it as resolved once they are satisfied. The developer should not mark questions raised by the reviewer as resolved since the reviewer uses these comments as a way to track what they are looking for in the next iteration of the review (tip: use an emoji such as the checkbox emoji to mark an inline comment as handled or addressed).
1. Reviewer re-reviews the code.
    1. result of the PR should be that the PR has moved to the next stage in the review ladder described above.
    1. this means that a PR will take a maximum of 2 review rounds to get to completion.
    1. back to the previous step for the author to make changes based on the re-review.

## Draft PRs

1. If there is a need to "try out the change" and get feedback, we can use a draft PR, although this should be used sparingly.
1. If a PR starts off in draft state then it should be used more as a "whiteboard for discussion" as opposed to the early stage of a PR.
1. In most cases, Draft PRs should be discarded. To avoid noise from comments, Draft PRs should be resubmitted as fresh standalone PRs instead of converting the draft PR to a regular PR.


glide.yaml
```diff
@@ -21,9 +21,6 @@ import:
   version: v1.0.0
 - package: github.com/form3tech-oss/jwt-go
   version: 5b2d2b5f6c34ccb3b6b65f77f4706558067690ef
-##############################################################
```
nikhilsaraf, line 24: Can you revert this comment (and associated closing comment below)?

glide.yaml
```diff
@@ -66,4 +60,8 @@ import:
 - package: github.com/denisbrodbeck/machineid
   version: v1.0.1
 - package: github.com/google/uuid
-  version: v1.1.2
\ No newline at end of file
+  version: v1.1.2
+- package: github.com/adshao/go-binance
+  version: ^2.3.0
```
nikhilsaraf, line 65: let's use the exact version `2.3.0` instead of `^2.3.0` for better predictability. we use that pattern for other dependencies as well.

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func isStale(data mapData, ttl time.Duration) bool {
+
+	return time.Now().Sub(data.createdAt).Seconds() > ttl.Seconds()
+}
+
+type mapEvents struct {
+	data map[string]mapData
+	mtx  *sync.Mutex
```
nikhilsaraf, line 83: if we can change this to a `*sync.RWMutex` that is ideal.
Then on Set we can do `rwlock.Lock()` and on Get() we can do `rwlock.RLock()`. This will be more performance than a mutex.

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+func (m *mapEvents) Set(key string, data interface{}) {
+
+	now := time.Now()
+
+	m.mtx.Lock()
+	m.data[key] = mapData{
+		data:      data,
+		createdAt: now,
+	}
+	m.mtx.Unlock()
```
nikhilsaraf, line 95: generally prefer using the defer pattern like this:
```golang
lock.Lock()
defer lock.Unlock()
```

this is a pattern we use across the codebase and is generally safer (without going into details here).

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+
+	return beWs, nil
+}
+
+func getPrecision(floatStr string) int8 {
+
+	strs := strings.Split(floatStr, ".")
+
+	if len(strs) != 2 {
+		log.Printf("error get precision for float %s\n", floatStr)
```
nikhilsaraf, line 189: if we are saying it is an error here we should panic. but it's ok to return 0 here and not an error.
can we change the message from "error get precision..." to "could not get precision..."?

this way when we see this log line we are just being informed of this rather than having consumers blaming this log line for some other actual error in the bot.

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+		return 0
+	}
+
+	return int8(len(strs[1]))
```
nikhilsaraf, line 193: 👍  nice!

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+
+		if !isTicker {
+			stream, err := subcribeTicker(symbol, beWs.events.SymbolStats)
+
+			if err != nil {
+				return nil, fmt.Errorf("error when subscribing for %s: %s", symbol, err)
+			}
+
+			//Store stream
+			beWs.mtx.Lock()
```
nikhilsaraf, line 218: can we call this `streamLock` instead of something generic like `mtx`?

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+			beWs.streams[symbol+STREAMTICKER] = stream
```
nikhilsaraf, line 219: for all streams, it would be better if we did this:
```golang
STREAM_TICKER_FMT = "%s@ticker"
beWs.streams[fmt.Sprintf(STREAM_TICKER_FMT, symbol)] = stream
```
that way it's very clear in the STREAM_TICKER_FMT variable where we will be inserting the variable. This also makes it a little easier to reference in the binance API.

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+			beWs.mtx.Unlock()
+
+			//Wait for binance to send events
+			time.Sleep(time.Second)
```
nikhilsaraf, line 223: can we pull this sleep time Duration as a const variable up? call it something like `waitForFirstEvent`
we can use this same value across all stream subscriptions.
In a future PR we should refactor the logic inside `if !isTicker {` so we can easily reuse -- but that's better handled in the next PR.

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+			BidPrice:  model.NumberFromFloat(bidPrice, getPrecision(ticker.BidPrice)),
+			LastPrice: model.NumberFromFloat(lastPrice, getPrecision(ticker.LastPrice)),
+		}
+	}
+
+	return priceResult, nil
+}
+
+//Unsubscribe ... unsubscribe from binance streams
+func (beWs *binanceExchangeWs) Unsubscribe(stream string) {
```
nikhilsaraf, line 274: 👍

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,283 @@
...
+
+	beWs.mtx.Lock()
+
+	if stream, isStream := beWs.streams[stream]; isStream {
+		stream.Close()
+	}
```
nikhilsaraf, line 280: do we also need to remove all values from `beWs.streams`?

plugins/binanceExchange_ws_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stellar/kelp/model"
+	"github.com/stretchr/testify/assert"
+)
+
+// import (
```
nikhilsaraf, line 11: can remove these commented out lines. don't need them in the final codebase

plugins/binanceExchange_ws_test.go
```diff
@@ -0,0 +1,71 @@
...
+// 	"github.com/stellar/kelp/api"
+// 	"github.com/stellar/kelp/model"
+// )
+
+var testOrderConstraintsBinanceWs = map[string]map[model.TradingPair]model.OrderConstraints{
+	"binance": {
+		*model.MakeTradingPair(model.XLM, model.USDT): *model.MakeOrderConstraints(4, 5, 0.1),
+		*model.MakeTradingPair(model.XLM, model.BTC):  *model.MakeOrderConstraints(8, 4, 1.0),
+	},
+	"kraken": {
```
nikhilsaraf, line 30: why do we have a reference to kraken in a test for binance exchange?
maybe we can make this map simpler by just being `map[model.TradingPair]model.OrderConstraints` instead of `map[string]map[model.TradingPair]model.OrderConstraints`?

I think this code is copied from ccxtExchange.go which is wrong. copying from krakenExchange.go is correct.

plugins/binanceExchange_ws_test.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func Test_createStateEvents(t *testing.T) {
+
+	events := createStateEvents()
+
+	assert.NotNil(t, events)
+}
+
+func Test_binanceExchangeWs_GetTickerPrice(t *testing.T) {
```
nikhilsaraf, line 46: 👍

nikhilsaraf (commented):
Added some comments inline. Should be good to merge after that and will set a good pattern for the rest of the PRs. First one is the hardest so thank you for getting through it!

nikhilsaraf:
Link to CLA: https://docs.google.com/forms/d/e/1FAIpQLSc5eppq6GOu9-TnFuRMBh4qIP1ChmZx9lrA6zOTyYiowKiwpA/viewform

plugins/binanceExchange_ws.go
tibrn: We can remove them, but it's only if we want this

glide.lock
```diff
@@ -41,8 +44,12 @@ imports:
   - url
 - name: github.com/asticode/go-bindata
   version: 5ad1fa75c07e881fbf05d8aa29bf88008c0da5ab
+- name: github.com/auth0/go-jwt-middleware
+  version: 1c6db3ceac5f4d28b35ee4b9949a3f80030aa6a3
 - name: github.com/Beldur/kraken-go-api-client
   version: 8d8ccfe4cc60d1703ffe596c889b887a318a884c
+- name: github.com/bitly/go-simplejson
```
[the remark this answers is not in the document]
tibrn: image
yes, it's from go-binance

plugins/binanceExchange_ws.go
tibrn: I had the same thought when I was working on the next milestone, I will add the refactor on the next PR

plugins/binanceExchange_ws.go
tibrn: https://github.com/golang/go/issues/14939, I can do that, but we have to take in consideration the performance issue from what I know this was improved in go 1.14+
image
https://golang.org/doc/go1.14, but the project is using go 1.13

glide.lock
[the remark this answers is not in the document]
nikhilsaraf: makes sense. thanks 👍

glide.lock
```diff
@@ -140,10 +151,6 @@ imports:
   - vfsutil
 - name: github.com/shurcooL/vfsgen
   version: 6a9ea43bcacdf716a5c1b38efff722c07adf0069
-- name: github.com/sirupsen/logrus
```
nikhilsaraf: we need this `github.com/sirupsen/logrus`
we don't need `github.com/Sirupsen/logrus`
the difference is the lowercase and upper case. Lowercase is required (this is an issue all go devs who use this library face because it did a weird rename between uppercase/lowercase)

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/stellar/kelp/model"
+)
+
+const (
+	STREAM_TICKER_FMT = "%s@ticker"
+	TTLTIME           = time.Second * 3 // ttl time in seconds
+)
+
+var (
+	waitForFirstEvent = true
```
nikhilsaraf: should be `time.Second`, see details in our chat

plugins/binanceExchange_ws.go
nikhilsaraf: yes lets' remove please

plugins/binanceExchange_ws.go
nikhilsaraf: discussed in our chat. `defer` for now since we will be upgrading beyond go 1.13 soon

plugins/binanceExchange_ws.go
```diff
@@ -215,12 +227,14 @@ func (beWs *binanceExchangeWs) GetTickerPrice(pairs []model.TradingPair) (map[mo
...
-			beWs.mtx.Unlock()
+			beWs.streamLock.Lock()
+			beWs.streams[fmt.Sprintf(STREAM_TICKER_FMT, symbol)] = stream
+			beWs.streamLock.Unlock()
 
-			//Wait for binance to send events
-			time.Sleep(time.Second)
+			if waitForFirstEvent {
+				//Wait for binance to send events
+				time.Sleep(time.Second)
```
nikhilsaraf: don't need `waitForFirstEvent` as a bool, it should be a time.Duration = time.Second.
then we can do this: `time.Sleep(waitForFirstEvent)`.

plugins/binanceExchange_ws.go
```diff
@@ -0,0 +1,299 @@
...
+
+	data, isData := m.data[key]
+
+	return data, isData
+}
+
+//Del ... delete cached value
+func (m *mapEvents) Del(key string) {
+	m.mtx.Lock()
+	m.mtx.Unlock()
```
nikhilsaraf, line 121: forgot the `defrer` keyword in `defer m.mtx.Unlock()`

nikhilsaraf (approved):
LGTM, thanks!

nikhilsaraf:
link to CLA: https://docs.google.com/forms/d/e/1FAIpQLSc5eppq6GOu9-TnFuRMBh4qIP1ChmZx9lrA6zOTyYiowKiwpA/viewform

PR #719, opened 2021-07-19
Feat/binance exchange ws get trade history/get latest trade cursor

tibrn:
# Guidelines for submitting code to Kelp

## Before Submitting a PR

1. Problem Identification
    1. ensure you understand the problem / task.
    1. ask initial questions to understand the task described in the Github issue, which may not be clearly defined.
    1. you can reach the team on keybase or by commenting on the Github issue.
    1. this is not the point to discuss concrete solutions (concrete design or implementation), but only what the problem is and the desired outcome.
    1. ask any questions related to scope of work, breakdown of tasks into subtasks and PRs.
    1. typically this should be covered in the Github issue. If it is not, then add to the issue after discussion with code owner.
    1. author needs to be comfortable with breakdown and scope of tasks. Scope should be small. Think about the overhead on the code reviewer when breaking down the tasks.

1. Research Solution
    1. conduct research on the issue at hand. Identify point of code change and any “fallout” from the change (tip: search the codebase for consumers of the function / object, and their consumers etc., until you get to a part of the code that you are well-versed with).
    1. author needs to become the “expert” on the part of the code being touched before submitting a code change.
    1. If it will take you 2 more days to figure something out by reading the code or experimenting with the codebase then do it.
        1. This is not wasted effort.
        1. It is likely that the code reviewer would use a similar process to figure it out, but this is your job and not the code reviewer’s job.
    1. If the process of becoming an “expert” on a specific area would take you more than 2 days then discuss this with the code reviewer if it is worth it, and proceed accordingly.
    1. Think about what kind of tests you would want. Are you going with a black-box testing approach, or a white-box testing approach? Why?
1. Technical Design Discussion
    1. PRs are not the best medium to conduct design discussions. The team is readily accessible and available so 1-1 meetings can be very effective.
    1. This is the best avenue to discuss your ideas for how to approach the problem.
    1. Prepare any relevant technical documentation for the design discussion and post on the Github issue and ping code owner.
        1. This can be in the form of a google document, flowchart diagram, or in rare cases a code snippet in a Draft PR.
        1. The Draft PR should not be reviewed by the code reviewer as part of the code change but can be used as a space to get feedback
        1. Resulting comments on a Draft PR can capture any guidance from the technical discussion.
        1. All Draft PRs should be discarded and never converted to a non-draft PR to avoid the overhead of comments lingering from the design discussion using the Draft PR.
    1. Schedule technical design discussion to go over the proposed solution.
    1. Include a list of tests you need to add, remove, modify. Consider all possibilities (tip: data related tests would require many test cases).
1. Implementation
    1. implement feature or bug fix. This should be the first line of code that you write for the PR.
    1. Avoid for loops to “autogenerate” test cases, unless absolutely necessary.
1. Pull Request
    1. The first version of the PR should not have any “surprises” compared to what was discussed in the technical design.
    1. See [Handling of PRs](#handling-of-prs) section below for guide on how to get a PR to merged status.
    1. See [detailed guide on submitting code-reviewer-friendly PRs](https://mtlynch.io/code-review-love/). This guide will save a lot of time for both author and reviewer and will help speed up the turnaround time on PRs.

## Handling of PRs

Typically, this is what is expected from the author of a Pull Request:

1. Each PR should correspond to the *smallest independent logical unit* of change.
    1. each PR should be prefixed with "bugfix", "feature", or "refactor".
    1. refactors should never be mixed in with logic changes or bug fixes and vice-versa.
    1. refactors that change variable names across more than 1 function should be a separate PR altogether.
1. Author should add inline commentary on the code changes to "guide" the reviewer on how to read the code change, highlighting key changes made.
    1. this should typically be 2-3 places, as anything bigger is too big of a PR and should be split up.
1. We should aim for the following benchmarks in terms of PR efficiency:
    1. In 60% of the cases the PR should be approved without any changes requested from the reviewer.
    1. In 25% of the cases the reviewer should add 2-3 comments that need to be "fixed up". This is most likely a code style suggestion.
    1. In 10% of the cases the reviewer should request a structural change to the code. These should not be "new ideas".
    1. In <=5% of the cases the PR may be discarded because we have learned something new in the process of implementation.
1. Author makes requested changes.
    1. no "new ideas" to be introduced at this stage.
    1. no structural changes to be introduced beyond what is requested.
    1. focus should be on minimizing "changes" beyond what is requested by reviewer.
    1. person who created a comment / raised a question on the PR is responsible to mark it as resolved once they are satisfied. The developer should not mark questions raised by the reviewer as resolved since the reviewer uses these comments as a way to track what they are looking for in the next iteration of the review (tip: use an emoji such as the checkbox emoji to mark an inline comment as handled or addressed).
1. Reviewer re-reviews the code.
    1. result of the PR should be that the PR has moved to the next stage in the review ladder described above.
    1. this means that a PR will take a maximum of 2 review rounds to get to completion.
    1. back to the previous step for the author to make changes based on the re-review.

## Draft PRs

1. If there is a need to "try out the change" and get feedback, we can use a draft PR, although this should be used sparingly.
1. If a PR starts off in draft state then it should be used more as a "whiteboard for discussion" as opposed to the early stage of a PR.
1. In most cases, Draft PRs should be discarded. To avoid noise from comments, Draft PRs should be resubmitted as fresh standalone PRs instead of converting the draft PR to a regular PR.

### DIFF
--- a/plugins/binanceExchange_ws.go
+++ b/plugins/binanceExchange_ws.go
@@ -278,11 +278,11 @@ func (beWs *binanceExchangeWs) subscribeStream(symbol, format string, subscribe 
 	//Wait for binance to send events
 	time.Sleep(timeToWaitForFirstEvent)
 
-	data, isStream := beWs.events.SymbolStats.Get(symbol)
+	data, isStream := state.Get(symbol)
 
 	//We couldn't subscribe for this pair
 	if !isStream {
-		return mapData{}, fmt.Errorf("error while fetching ticker price for trading pair %s", symbol)
+		return mapData{}, fmt.Errorf("error while subscribing for %s", fmt.Sprintf(format, symbol))
 	}
 
 	return data, nil
@@ -354,8 +354,7 @@ func (beWs *binanceExchangeWs) GetTickerPrice(pairs []model.TradingPair) (map[mo
 func (beWs *binanceExchangeWs) GetOrderBook(pair *model.TradingPair, maxCount int32) (*model.OrderBook, error) {
 
 	var (
-		maxCountInt = int(maxCount)
-		fetchLimit  = maxCountInt
+		fetchLimit = int(maxCount)
 	)
 
 	if fetchLimit > 20 {
@@ -397,6 +396,12 @@ func (beWs *binanceExchangeWs) GetOrderBook(pair *model.TradingPair, maxCount in
 
 	askCcxtOrders := book.Asks
 	bidCcxtOrders := book.Bids
+
+	if fetchLimit != 20 {
+		// we may not have fetched all the requested levels because the exchange may not have had that many levels in depth
+		askCcxtOrders = askCcxtOrders[:fetchLimit]
+		bidCcxtOrders = bidCcxtOrders[:fetchLimit]
+	}
 
 	asks, err := beWs.readOrders(askCcxtOrders, pair, model.OrderActionSell)
 
--- a/plugins/binanceExchange_ws.go
+++ b/plugins/binanceExchange_ws.go
@@ -9,12 +9,14 @@ import (
 	"time"
 
 	"github.com/adshao/go-binance/v2"
+	"github.com/adshao/go-binance/v2/common"
 	"github.com/stellar/kelp/api"
 	"github.com/stellar/kelp/model"
 )
 
 const (
 	STREAM_TICKER_FMT = "%s@ticker"
+	STREAM_BOOK_FMT   = "%s@depth"
 	TTLTIME           = time.Second * 3 // ttl time in seconds
 )
 
@@ -23,9 +25,11 @@ var (
 )
 
 var (
-	ErrConversionWsMarketEvent = errConversion{from: "interface", to: "*binance.WsMarketStatEvent"}
+	ErrConversionWsMarketEvent       = errConversion{from: "interface", to: "*binance.WsMarketStatEvent"}
+	ErrConversionWsPartialDepthEvent = errConversion{from: "interface", to: "*binance.WsPartialDepthEvent"}
 )
 
+type Subscriber func(symbol string, state *mapEvents) (*stream, error)
 type errMissingSymbol struct {
 	symbol string
 }
@@ -135,17 +139,21 @@ func makeMapEvents() *mapEvents {
 //struct used to keep all cached data
 type events struct {
 	SymbolStats *mapEvents
+	BookStats   *mapEvents
 }
 
 func createStateEvents() *events {
 	events := &events{
 		SymbolStats: makeMapEvents(),
+		BookStats:   makeMapEvents(),
 	}
 
 	return events
 }
 
-// subscribe for symbol@ticker
+// 24hr rolling window ticker statistics for a single symbol. These are NOT the statistics of the UTC day, but a 24hr rolling window for the previous 24hrs.
+// Stream Name: <symbol>@ticker
+// Update Speed: 1000ms
 func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
 
 	wsMarketStatHandler := func(ticker *binance.WsMarketStatEvent) {
@@ -161,6 +169,49 @@ func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	keepConnection(doneC, func() {
+		subcribeTicker(symbol, state)
+	})
+
+	return &stream{doneC: doneC, stopC: stopC, cleanup: func() {
+		state.Del(symbol)
+	}}, err
+
+}
+
+//restart Connection with ws// Binance close each connection after 24 hours
+func keepConnection(doneC chan struct{}, reconnect func()) {
+
+	go func() {
+		<-doneC
+		reconnect()
+	}()
+}
+
+// Top <levels> bids and asks, pushed every second. Valid <levels> are 5, 10, or 20.
+// <symbol>@depth<levels>@100ms
+// 100ms
+func subcribeBook(symbol string, state *mapEvents) (*stream, error) {
+
+	wsPartialDepthHandler := func(event *binance.WsPartialDepthEvent) {
+		state.Set(symbol, event)
+	}
+
+	errHandler := func(err error) {
+		log.Printf("Error WsPartialDepthServe for symbol %s: %v\n", symbol, err)
+	}
+
+	//Subscribe to highest level
+	doneC, stopC, err := binance.WsPartialDepthServe100Ms(symbol, "20", wsPartialDepthHandler, errHandler)
+
+	if err != nil {
+		return nil, err
+	}
+
+	keepConnection(doneC, func() {
+		subcribeBook(symbol, state)
+	})
 
 	return &stream{doneC: doneC, stopC: stopC, cleanup: func() {
 		state.Del(symbol)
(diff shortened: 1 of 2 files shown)


plugins/binanceExchange_ws_test.go
```diff
@@ -89,3 +103,70 @@ func Test_binanceExchangeWs_GetOrderBook(t *testing.T) {
...
+
+func Test_binanceExchangeWs_GetLatestTradeCursor(t *testing.T) {
+
+	if testing.Short() {
+		return
+	}
+
+	startIntervalSecs := time.Now().Unix()
+
+	testBinanceExchangeWs, err := makeBinanceWs(emptyAPIKeyBinance)
```
nikhilsaraf, line 115: does this test pass without using valid API keys? -- I think this test should fail without valid keys but I'm not sure (see comment above about throwing an error when we can't open up the cursor stream)

plugins/binanceExchange_ws_test.go
```diff
@@ -89,3 +103,70 @@ func Test_binanceExchangeWs_GetOrderBook(t *testing.T) {
...
+		return
+	}
+	if !assert.True(t, endIntervalSecs >= cursorInt, fmt.Sprintf("returned cursor (%d) should be lte the end time of the function call in millis (%d)", cursorInt, endIntervalSecs)) {
+		return
+	}
+}
+
+func Test_binanceExchangeWs_GetTradeHistory(t *testing.T) {
+
+	if testing.Short() {
```
nikhilsaraf, line 147: makes sense to put this here because we can't run this in the CI pipeline since it needs API keys

were you able to test this with your own binance API keys?

plugins/binanceExchange_ws.go
```diff
@@ -180,6 +262,98 @@ func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
...
+	}
+
+	errHandler := func(err error) {
+		log.Printf("Error WsUserDataServe for listenKey %s: %v\n", listenKey, err)
+	}
+
+	doneC, stopC, err := binance.WsUserDataServe(listenKey, wsUserStreamHandler, errHandler)
+
+	if err != nil {
+		return nil, err
```
nikhilsaraf, line 344: can we wrap this error please (see comments about error wrapping below)?

plugins/binanceExchange_ws.go
```diff
@@ -15,9 +18,12 @@ import (
 )
 
 const (
-	STREAM_TICKER_FMT = "%s@ticker"
-	STREAM_BOOK_FMT   = "%s@depth"
-	TTLTIME           = time.Second * 3 // ttl time in seconds
+	STREAM_TICKER_FMT      = "%s@ticker"
+	STREAM_BOOK_FMT        = "%s@depth"
+	STREAM_USER            = "@user"
```
nikhilsaraf, line 23: `@user` is just something we made up correct? it's not actually part of the binance spec right?

If so, can you please add a comment specifying that this is not used by binance and just a key we use to reference into the `streams` map?

(we should do the same for `LAST_CURSOR_KEY` below if that is also just a key that we made up)

plugins/binanceExchange_ws.go
```diff
@@ -15,9 +18,12 @@ import (
...
+	LAST_CURSOR_KEY        = "@user||lastCursor"
```
nikhilsaraf, line 24: same as above for `LAST_CURSOR_KEY`

plugins/binanceExchange_ws.go
```diff
@@ -180,6 +262,98 @@ func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
...
+
+	userStreamLock := sync.Mutex{}
+
+	wsUserStreamExecutinReportHandler := func(message []byte) {
+
+		event := &eventExecutionReport{}
+		err := json.Unmarshal(message, event)
+
+		if err != nil {
+			log.Printf("Error unmarshal %s to eventExecutionReport\n", string(message))
```
nikhilsaraf, line 275: If this fails here then does that mean that `subcribeUserStream` has failed?

If so then we should also cause `subcribeUserStream` to return an error in this case.

plugins/binanceExchange_ws.go
```diff
@@ -180,6 +262,98 @@ func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
...
+			return
+		}
+
+		userStreamLock.Lock()
```
nikhilsaraf, line 279: why do we need to lock here?
wondering in which scenario may this be executed twice, or maybe I've missed something

plugins/binanceExchange_ws.go
```diff
@@ -180,6 +262,98 @@ func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
...
+			state.Set(event.Symbol, history)
+		}
+
+		now := time.Now()
+		history.createdAt = now
+
+		data, isOk := history.data.(History)
+
+		if !isOk {
+			log.Printf("Error conversion %v\n", ErrConversionHistory)
```
nikhilsaraf, line 295: should this also cause the stream to fail and throw an error somewhere?

I'm concerned that all these errors here will be hidden errors that will not cause the code to crash.

we want the bot to crash rather than throw silent errors

plugins/binanceExchange_ws.go
```diff
@@ -180,6 +262,98 @@ func subcribeTicker(symbol string, state *mapEvents) (*stream, error) {
...
+		if isCursor {
+
+			cursor, isOk := lastCursorData.data.(int64)
+
+			if isOk {
+				if cursor > lastCursor {
+					lastCursor = cursor
+				}
+			} else {
+				log.Printf("Error converting cursor %v\n", ErrConversionCursor)
```
nikhilsaraf, line 314: maybe we should return an error here rather than just logging. we want the bot to crash if there is any error

plugins/binanceExchange_ws.go
```diff
@@ -140,12 +220,14 @@ func makeMapEvents() *mapEvents {
 type events struct {
 	SymbolStats *mapEvents
 	BookStats   *mapEvents
+	OrderEvents *mapEvents
```
nikhilsaraf, line 223: Why is this called `OrderEvents`?

I'm confused because we are fetching the User Stream and getting Trade History, so was expecting this to be called either `TradeHistoryEvents` or `UserEvents`

nikhilsaraf (commented):
added some comments. this one turned out to be a bit heavier than the first two.

plugins/binanceExchange_ws.go
tibrn: the only way to capture errors when binance it's sending events it's to use another map(state) to save the error when happens and when any method from the interface it's called to check if for the stream "X" which it's using is any error

plugins/binanceExchange_ws.go
tibrn: this is to avoid any misusage of the callback and to not have invalid data, because I am modifying the slice in the callback and the possibility for that callback to be executed by two goroutines exists I've added the mutex

plugins/binanceExchange_ws.go
tibrn: one option it's to panic when I get errors from the stream

plugins/binanceExchange_ws.go
tibrn: oke, I will return an error for conversion

plugins/binanceExchange_ws.go
```diff
@@ -220,6 +394,19 @@ func subcribeBook(symbol string, state *mapEvents) (*stream, error) {
 
 }
 
+//ListenKey expires every 60 minutes
+func keepAliveStreamService(client *binance.Client, key string) {
```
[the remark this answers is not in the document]
tibrn: yes

plugins/binanceExchange_ws_test.go
```diff
@@ -16,10 +21,15 @@ func Test_createStateEvents(t *testing.T) {
 }
 
 func Test_binanceExchangeWs_GetTickerPrice(t *testing.T) {
+
+	if testing.Short() {
```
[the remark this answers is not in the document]
tibrn: Sure, I will see how I can make some refactor to use these tests in CI pipeline

plugins/binanceExchange_ws.go
nikhilsaraf: If the stream errors in update 1, but passes in update 2, and the user fetched the value of update 2 only, we don’t care about the error in update 1.

With that in mind, what I’m thinking is that when we write a value to the existing map (the data structure), we also include an error. So if there is an error in the stream we should write an error value.

When the user fetched the value they will check the error first and if it is an error they will handle appropriately (bubble upstream IIRC).


That allows us to keep the date of the response and error encapsulated into one map. What do you think?

plugins/binanceExchange_ws.go
nikhilsaraf: Ok will look at this a little more closely when I’m on a computer

plugins/binanceExchange_ws.go
nikhilsaraf: See above comment. Once we agree on a pattern we can apply to all streams.

plugins/binanceExchange_ws.go
```diff
@@ -220,6 +327,20 @@ func subcribeBook(symbol string, state *mapEvents) (*stream, error) {
...
+
+	for {
+		time.Sleep(time.Minute * 50)
+		err := client.NewKeepaliveUserStreamService().ListenKey(key).Do(context.Background())
+
+		if err != nil {
+			log.Printf("Error keepAliveStreamService %v\n", err)
+			panic(err)
```
nikhilsaraf, line 339: We should find a way to not panic here. Sorry didn’t see this earlier.
Panics are very bad in kelp for many reasons and I can go into detail if interested.

PR #723, opened 2021-07-24
feat(277): upgrade dependency management go modules

clD11:
This PR migrates Kelp from using glide to go modules for our dependency management

* Glide has been removed
* go.mod added
* Readme updated
* CircleCI updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,9 +7,9 @@ commands:
           keys:
             - go-mod-v4-{{ checksum "go.sum" }}
             - v5-node-cache
-      - run:
-          name: Fetch astilectron-bundler
-          command: go get -u github.com/asticode/go-astilectron-bundler/...
+#      - run:
+#          name: Fetch astilectron-bundler
+#          command: go get -u github.com/asticode/go-astilectron-bundler/...
       - run:
           name: Install astilectron-bundler
           command: go install github.com/asticode/go-astilectron-bundler
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,17 +7,17 @@ commands:
           keys:
             - go-mod-v4-{{ checksum "go.sum" }}
             - v5-node-cache
+      - run:
+          name: Download Modules
+          command: |
+            export GOPROXY=https://goproxy.io,https://proxy.golang.org,https://goproxy.cn
+            go mod download
 #      - run:
 #          name: Fetch astilectron-bundler
 #          command: go get -u github.com/asticode/go-astilectron-bundler/...
       - run:
           name: Install astilectron-bundler
           command: go install github.com/asticode/go-astilectron-bundler
-      - run:
-          name: Download Modules
-          command: |
-            export GOPROXY=https://goproxy.io,https://proxy.golang.org,https://goproxy.cn
-            go mod download
       - run:
           name: Install nvm
           command: curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,10 +9,10 @@ commands:
             - v5-node-cache
       - run:
           name: Fetch astilectron-bundler
-          command: go get github.com/asticode/go-astilectron-bundler@2c8a7b06a9020f358a06ff377331d0dd690de6de
+          command: go get github.com/asticode/go-astilectron-bundler\...
       - run:
           name: Install astilectron-bundler
-          command: go install github.com/asticode/go-astilectron-bundler@2c8a7b06a9020f358a06ff377331d0dd690de6de
+          command: go install github.com/asticode/go-astilectron-bundler
       - run:
           name: Download Modules
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,6 @@ commands:
           name: Download Go modules
           command: |
             export GOPROXY=https://goproxy.io,https://proxy.golang.org,https://goproxy.cn,direct
-            echo $GOPROXY
             go mod download
       - run:
           name: Fetch astilectron-bundler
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -124,25 +124,17 @@ _Note for Windows Users: You should use a [Bash Shell][bash] to follow the steps
...
-    * Change to `kelp` directory for all following commands: `cd $GOPATH/src/github.com/stellar/kelp`
-5. Install the dependencies:
-    * `glide install`
-6. Install the [astilectron-bundler][astilectron-bundler] binary into `$GOBIN`
-    * `go get -u github.com/asticode/go-astilectron-bundler/...`
+1. [Download][golang-download] and [setup][golang-setup] Golang _v1.13 or later_
+   1. Set environment variable `export GOPROXY=https://proxy.golang.org`
+2. Install [Yarn][yarn-install] and [NodeJs][nodejs-install] to build the Kelp GUI
+3. Clone the kelp repository `git clone git@github.com:stellar/kelp.git`
+4. Install the [astilectron-bundler][astilectron-bundler] binary into `$GOPATH/bin`
```
nikhilsaraf, line 131: We do not define where to set GOPATH, do we need to or can we remove it?

.circleci/config.yml
```diff
@@ -5,20 +5,17 @@ commands:
...
-          command: sudo apt-get update -y && sudo apt-get install -y golang-glide
-      - run:
-          name: Install dependencies using glide
-          command: glide install
-      - run:
-          name: Fetch astilectron-bundler
-          command: go get -u github.com/asticode/go-astilectron-bundler/...
+          name: Download Go Modules
+          command: |
+            export GOPROXY=https://goproxy.io,https://proxy.golang.org,https://goproxy.cn
```
nikhilsaraf, line 13: We should test only with proxy.Golang.org

.circleci/config.yml
```diff
@@ -5,20 +5,17 @@ commands:
...
-          name: Install glide
-          command: sudo apt-get update -y && sudo apt-get install -y golang-glide
-      - run:
-          name: Install dependencies using glide
-          command: glide install
-      - run:
-          name: Fetch astilectron-bundler
-          command: go get -u github.com/asticode/go-astilectron-bundler/...
+          name: Download Go Modules
+          command: |
```
nikhilsaraf, line 12: Why start out these commands with |
any advantage?

scripts/build.sh
```diff
@@ -120,7 +122,7 @@ function download_file() {
     echo "... downloaded $NAMESPACE to $DESTINATION"
 }
 
-if [[ $(basename $("pwd")) != "kelp" ]]
+if [[ $(basename $("pwd")) != "kelp" && $("pwd") != *"circleci/project"* ]];
```
nikhilsaraf, line 125: Why add the circleci exception now when it wasn’t needed earlier?

plugins/ccxtExchange_test.go
```diff
@@ -556,24 +557,17 @@ func TestCancelOrder_Ccxt(t *testing.T) {
 }
 
 func TestGetOrderConstraints_Ccxt_Precision(t *testing.T) {
-	// coinbasepro gives incorrect precision values so we do not test it here
```
nikhilsaraf, line 559: Why modify this file?
(Wondering how it’s related to the change)

go.mod
```diff
@@ -0,0 +1,68 @@
+module github.com/stellar/kelp
+
+go 1.13
+
+require (
+	cloud.google.com/go v0.84.0 // indirect
+	github.com/Beldur/kraken-go-api-client v0.0.0-20180126083054-8d8ccfe4cc60
```
nikhilsaraf, line 7: Seems reasonable — any dependency version updated here or it’s a direct port of the previous versions?

nikhilsaraf (commented):
Some questions. Seems good besides that

.circleci/config.yml
clD11: I tried that yesterday and it failed so had to add the others, I think its actually this https://goproxy.cn one that has that lib but kept the others for completeness.

go.mod
clD11: Direct port all the versions should be the same

README.md
clD11: I think GOPATH is synonymous with wherever your go workspace is as they will still need to  install astilectron-bundler in either where ever they set GOPATH or default $HOME/go

clD11: I need to update the proxy value though

.circleci/config.yml
clD11: Just when executing multiple individual  commands  over multiple lines I think its cleaner than > or using sh

scripts/build.sh
clD11: There was a very specific reason for this I think its the way the modules install and the new directory structure but it was ages ago and I have forgotten the details :laughing:

plugins/ccxtExchange_test.go
clD11: just to get the build green

README.md
nikhilsaraf: Many devs are new to Golang and will get confused by the GOPATH reference here. Go modules doesn’t need GOPATH anymore so we can just suggest that they install kelp into ~/kelp and use that as the path references throughout so it is simple for the reader who is new to Golang

.circleci/config.yml
nikhilsaraf: Ok then let’s include all 3 in the readme too so it is consistent with the CI setup

.circleci/config.yml
nikhilsaraf: Ok fair

scripts/build.sh
nikhilsaraf: Ok no worries. Let’s leave it in, if you remember the reason then please add that context as a comment next to the code for posterity :)

plugins/ccxtExchange_test.go
nikhilsaraf: Once we merge this fix we can rebase and this will work.

nikhilsaraf (approved):
Lgtm, thanks!

PR #728, opened 2021-09-13
feat(424): generic price feed

clD11:
This PR adds a generic price feed which takes a url and response path separated by a semi-colon to retrieve the price

Two examples urls and responses paths are shown below, to see the expected json response paste the url in the browser without the response path. This feat uses a library gjson and the full syntax for traversing json responses can be seen here https://github.com/tidwall/gjson/blob/master/SYNTAX.md

Examples

https://api.kraken.com/0/public/Ticker?pair=XBTUSD;result.XXBTZUSD.a.0
https://api.kraken.com/0/public/Depth?pair=XBTUSD;result.XXBTZUSD.asks.0.0

### DIFF
--- a/plugins/priceFeed.go
+++ b/plugins/priceFeed.go
@@ -2,6 +2,8 @@ package plugins
 
 import (
 	"fmt"
+	"github.com/stellar/kelp/support/json"
+	networking "github.com/stellar/kelp/support/networking"
 	"strings"
 
 	"github.com/stellar/go/clients/horizonclient"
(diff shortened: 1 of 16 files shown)


plugins/priceFeed.go
```diff
@@ -42,6 +44,13 @@ func MakePriceFeed(feedType string, url string) (api.PriceFeed, error) {
 		return newFiatFeed(url), nil
 	case "fiat-oxr":
 		return newFiatFeedOxr(url), nil
+	case "generic-price-feed":
+		parts := strings.Split(url, ";")
```
nikhilsaraf, line 48: Can we move the extraction of the url parts into the factory method. So it should just be one call:

makeGenericPriceFeed(url)

like we do for the other feeds. What do you think?

nikhilsaraf: Logic for delimiter seems reasonable

support/networking/client.go
```diff
@@ -0,0 +1,35 @@
+package networking
+
```
nikhilsaraf, line 2: What is the idea behind using this new class?

nikhilsaraf (commented):
Some comments. Should be good to go

will review in depth tomorrow

support/networking/client.go
clD11: just  so we have a single client and can pass this as a dependency into other functions/structs that need an http client we now just need to test it once. Also we are separating the more low level calls from say the price feed and following single responsibility

PR #729, opened 2021-09-13
fix broken precision value test

clD11:
This PR fixes the broken precision value test

### DIFF
--- a/plugins/ccxtExchange_test.go
+++ b/plugins/ccxtExchange_test.go
@@ -2,6 +2,7 @@ package plugins
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"log"
 	"math"
 	"strconv"
@@ -572,8 +573,8 @@ func TestGetOrderConstraints_Ccxt_Precision(t *testing.T) {
 			// }, {
 			exchangeName:       "binance",
 			pair:               &model.TradingPair{Base: model.XLM, Quote: model.USDT},
-			wantPricePrecision: 5,
-			wantVolPrecision:   1,
+			wantPricePrecision: 4,
+			wantVolPrecision:   4,
 		}, {
 			exchangeName:       "binance",
 			pair:               &model.TradingPair{Base: model.XLM, Quote: model.BTC},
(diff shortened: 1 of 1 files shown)


plugins/ccxtExchange_test.go
```diff
@@ -556,24 +557,17 @@ func TestCancelOrder_Ccxt(t *testing.T) {
 }
 
 func TestGetOrderConstraints_Ccxt_Precision(t *testing.T) {
-	// coinbasepro gives incorrect precision values so we do not test it here
```
nikhilsaraf, line 559: Can we leave this comment in here

plugins/ccxtExchange_test.go
```diff
@@ -556,24 +557,17 @@ func TestCancelOrder_Ccxt(t *testing.T) {
...
 	testCases := []struct {
 		exchangeName       string
 		pair               *model.TradingPair
 		wantPricePrecision int8
 		wantVolPrecision   int8
 	}{
 		{
-			// disable ccxt-kraken based tests for now because of the 403 Forbidden Security check API error
```
nikhilsaraf, line 567: Let’s leave this comment in here too. It’s a TODO for me but I forgot to mark as such.
Can prefix with // TODO NS

plugins/ccxtExchange_test.go
```diff
@@ -587,28 +581,20 @@ func TestGetOrderConstraints_Ccxt_Precision(t *testing.T) {
...
-				[]api.ExchangeAPIKey{emptyAPIKey},
-				[]api.ExchangeParam{emptyParams},
-				[]api.ExchangeHeader{},
-				false,
-				getEsParamFactory(kase.exchangeName),
-			)
-			if !assert.NoError(t, e) {
-				return
-			}
+	for _, tc := range testCases {
```
nikhilsaraf, line 584: Can we leave the var as kase to keep consistent with rest of codebase?
Tc is probably a better abbreviation/var-name but can migrate to that later in one sweeping change

plugins/ccxtExchange_test.go
```diff
@@ -587,28 +581,20 @@ func TestGetOrderConstraints_Ccxt_Precision(t *testing.T) {
...
-			}
-			if !assert.Equal(t, kase.wantVolPrecision, result.VolumePrecision) {
-				return
-			}
+			epf := getEsParamFactory(tc.exchangeName)
+
+			ccxtExchange, e := makeCcxtExchange(tc.exchangeName, nil, []api.ExchangeAPIKey{emptyAPIKey},
+				[]api.ExchangeParam{emptyParams}, []api.ExchangeHeader{}, false, epf)
+
+			require.NoError(t, e)
```
nikhilsaraf, line 592: Good idea to use require, it’s exactly the behavior we want. We should migrate to using require in other places too long-term. I’ll do that at some point

plugins/ccxtExchange_test.go
```diff
@@ -556,24 +557,17 @@ func TestCancelOrder_Ccxt(t *testing.T) {
...
-			// 	pair:               &model.TradingPair{Base: model.XLM, Quote: model.USD},
-			// 	wantPricePrecision: 6,
-			// 	wantVolPrecision:   8,
-			// }, {
 			exchangeName:       "binance",
 			pair:               &model.TradingPair{Base: model.XLM, Quote: model.USDT},
-			wantPricePrecision: 5,
-			wantVolPrecision:   1,
+			wantPricePrecision: 4,
+			wantVolPrecision:   4,
```
nikhilsaraf, line 570: Can you double check volume precision, I think it should be 0, not 4

nikhilsaraf (commented):
Comments inline

plugins/ccxtExchange_test.go
clD11: there is some logic that sets it to the price when its 0 
	volumePrecision := ccxtMarket.Precision.Amount
	if volumePrecision == 0 {
		volumePrecision = ccxtMarket.Precision.Price
	}
	
	so it ends up at 4

nikhilsaraf (approved):
Lgtm, thanks

nikhilsaraf:
Failing because of glide so bypassing CI to merge. We are fixing glide issue by migrating to go modules in a separate PR

PR #748, opened 2021-11-04
Upgrade to use stellar/go@horizonclient-v8.0.0

### LINKED ISSUE #741: Fill Tracker cannot unmarshal string into Go struct field Price._embedded.records.price.n of type int32
## Describe the bug
Getting the following error in the latest build of Kelp when trying to fetch trades:
```
error when running an iteration of fill tracker: error when fetching trades: error while fetching trades in SDEX (cursor=???????????????-?): json: cannot unmarshal string into Go struct field Price._embedded.records.price.n of type int32
```

* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * 
**Please provide us with the relevant information in this [quick form (1 minute)](https://forms.gle/G1Ud6HZjubaP6qS1A) so we can track who is being faced with an outage on Kelp because of the Protocol 18 upgrade, and can help you get back online as quickly as possible
https://forms.gle/G1Ud6HZjubaP6qS1A**
* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * 

## Expected behavior
I think the correct behavior in this situation is for the bot to parse trade values (and price) correctly

## Frequency
The frequency is: since at least `2021/10/06 17:48:15 UTC`

## Steps To Reproduce
1. Run Kelp on SDEX with fill tracking enabled.
2. Execute a trade against your bot.
3. When the fill tracker runs, it will try and parse the trade and will fail with the above error

## Possible Solution
This bug can be fixed by patching the price parsing logic by either upgrading the library or running a parallel library import with the fix (since we are using an old version of the SDK)

## Your Environment
All versions of Kelp as off October 6th 2021

## Context
I am unable to run any trading template that uses the fill tracker such as the `balanced` or `mirror` templates because of this bug.

## Attachments
None

## References
See [this PR 3915 in the stellar/go repo](https://github.com/stellar/go/pull/3915) for the change made to horizon
need to switch over from using `Price` to using `TradePrice`

See updated Go SDK: https://github.com/stellar/go/releases/tag/horizonclient-v8.0.0-beta.0

Version of horizon released: https://github.com/stellar/go/releases/tag/horizon-v2.9.0rc1

leighmcculloch:
This PR is a draft and is likely incomplete.

I've taken a time-boxed pass at upgrading the `stellar/go` dependency from `horizonclient-v3.0.0` to `horizonclient-v8.0.0`. I mostly just wanted to explore, understand, and validate assumptions about how Kelp could be upgraded, assuming that the upgrade path would not be smooth.

The intention was to understand, not necessarily deliver an upgrade.

I've opened this draft PR to have this change run on Kelp's CI, to see if at least from CI's perspective if this idea for how to make the upgrade happen, has substance or creates more problems than it solves.

The approach I've taken is to:
- Upgrade `github.com/stellar/go`.
- Preserve use of the old `github.com/stellar/go@horizonclient-v3.0.0/build` by vendoring that package inside of kelp.
- Update references to types and functions within `github.com/stellar/go/xdr` to their new names, since some XDR types  had their names changed.
- Make zero functional changes.

Maybe fix #741


plugins/sdex.go
```diff
@@ -856,7 +857,11 @@ func (sdex *SDEX) tradesPage2TradeHistoryResult(baseAsset hProtocol.Asset, quote
 		if e != nil {
 			return nil, false, fmt.Errorf("could not convert baseAmount to model.Number: %s", e)
 		}
-		floatPrice := float64(t.Price.N) / float64(t.Price.D)
+		floatPrice, exact := big.NewRat(t.Price.N, t.Price.D).Float64()
+		if !exact {
+			// TODO: What needs to happen if the float price cannot be determined?
+			panic(fmt.Errorf("price %d / %d cannot be captured as a float64", t.Price.N, t.Price.D))
+		}
```
leighmcculloch, line 864: @nikhilsaraf This part of the change needs further investigation.

ddombrowsky (approved):
It's a bit of a hack to put in a module from the old version of the SDK, but I think it works for now.  In the long run it's cleaner than forking the SDK itself.

nikhilsaraf (approved):
LGTM, thanks @leighmcculloch !

PR #757, opened 2022-02-11
Make local development and running simpler

leighmcculloch:
### What
Don't require build information when running locally during development.

### Why
When developing locally it is useful to be able to run kelp as is with simple commands like `go run .`. That isn't possible today because kelp requires a variety of build parameters to be specified. That's inconvenient for local development, and it is only really necessary to have all these build parameters included when releasing.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -62,6 +62,8 @@ var RootCmd = &cobra.Command{
 			if e != nil {
 				panic(e)
 			}
+		} else {
+			panic(fmt.Sprintf("unrecognized buildType: %s", buildType))
 		}
 	},
 }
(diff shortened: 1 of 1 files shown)


ire-and-curses (approved):

nikhilsaraf (approved):

leighmcculloch:
@nikhilsaraf The integration test is failing with this error:
>2022/02/11 01:31:56 could not get last trade cursor from exchangeShim: error while fetching latest trade cursor in SDEX: horizon error: "Resource Missing" - check horizon.Error.Problem for more information (quoteAssetError: asset 'CITEST:GBMMZMK2DC4FFP4CAI6KCVNCQ7WLO5A7DQU7EC7WGHRDQBZB763X4OQI' did not exist)
2022/02/11 01:31:56 metric - not sending event metric of type 'delete_offers' because metrics are disabled

I don't think that is a result of the changes I'm making here. How do we fix this?

nikhilsaraf:
The secret keys for that are in the circle ci file. I manually initialize it and issue assets to the trader account when the testnet is reset

